PR #10, opened 2022-05-18
Sync from monorepo

cihati:


dpp23:
@cihati I think the previous run of the tests hit https://github.com/layerai/sdk/pull/13  so I cancelled and re-run as they seemed stuck.

Can you please update the poetry.lock since you edited `pyproject.toml`? I've added https://github.com/layerai/sdk/pull/12 to make sure we don't forget in the future.

cihati:
@dpp23 do you do this via `poetry update`? It's been running for 6 mins now and still haven't finalized, just wanted to make sure 😅

dpp23:
> @dpp23 do you do this via `poetry update`? It's been running for 6 mins now and still haven't finalized, just wanted to make sure sweat_smile

`poetry lock --no-update`. It takes like 40s for me.

PR #11, opened 2022-05-18
Dimitar/datadog

dpp23:


dpp23:
@cihati I think the previous run of the tests hit https://github.com/layerai/sdk/pull/13  so I cancelled and re-run as they seemed stuck.

Can you please update the poetry.lock since you edited `pyproject.toml`? I've added https://github.com/layerai/sdk/pull/12 to make sure we don't forget in the future.

PR #21, opened 2022-05-19
Increase retry timeout to 30s

volkangurel:
Increase retry timeout to 30s to improve reliability of long running jobs

### DIFF
--- a/layer/utils/grpc/channel.py
+++ b/layer/utils/grpc/channel.py
@@ -38,7 +38,7 @@ def create_grpc_channel(
                     "retryPolicy": {
                         "maxAttempts": 5,
                         "initialBackoff": "0.1s",
-                        "maxBackoff": "5s",
+                        "maxBackoff": "30s",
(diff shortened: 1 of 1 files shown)


layer/utils/grpc/channel.py
```diff
@@ -38,7 +38,7 @@ def create_grpc_channel(
                     "retryPolicy": {
                         "maxAttempts": 5,
                         "initialBackoff": "0.1s",
-                        "maxBackoff": "5s",
+                        "maxBackoff": "30s",
```
dpp23, line 41: I think you should increase the attempts not the backoff? I don't think we ever reach that now anyway

dpp23: We are going to do 0.1, 0.2, 0.4, 0.8, 1.6 and then exit with or without this change.

volkangurel: You're right, fixing

eminn (approved):

PR #26, opened 2022-05-20
Add pylint

dpp23:
Best to review commit by commit:

1. First commit introduces pylint
2. Second commit fixes linting issues


volkangurel (approved):

layer/decorators/layer_wrapper.py
```diff
@@ -12,3 +12,12 @@ def __init__(self, wrapped: Any, wrapper: Any, enabled: Any) -> None:
...
         return type(self), (self.__wrapped__, self._self_wrapper, self._self_enabled)
+
+    def __copy__(self) -> None:
+        pass
+
+    def __deepcopy__(self, memo: Dict[int, object]) -> None:
+        pass
+
+    def __reduce__(self) -> Union[str, Tuple[Any, ...]]:
+        pass
```
cihati, line 23: Why do we have to put this?

cihati (approved):
🎖️

layer/decorators/layer_wrapper.py
dpp23: Pylint was complaining that these are not implemented but marked as `abstractmethod` in the parent class.

PR #33, opened 2022-05-21
LAY-3102 - Fix bandit exclusions

dpp23:


pyproject.toml
```diff
@@ -305,3 +305,14 @@ enable = [
...
+
+[tool.bandit]
+exclude_dirs = [
+    "test",
+    ".eggs"
+]
+skips = [
+    "B101", # assert_used
+    "B105", # hardcoded_password_string too noisy on defaults values empty string
+    "B106", # hardcoded_password_funcarg too noisy on defaults values empty string
```
dpp23, line 317: These two fire any variable that has `token` in the name. We have a few places where we set empty string as opposed to None in order not to propagate None checks all over the place. It is debatable whether this is good code, but quite involved to fix.

pyproject.toml
```diff
@@ -305,3 +305,14 @@ enable = [
...
     "invalid-name"
 ]
+
+[tool.bandit]
+exclude_dirs = [
+    "test",
+    ".eggs"
+]
+skips = [
+    "B101", # assert_used
```
dpp23, line 315: We use `assert` pervasively in order to convince the typechecker that an `Optional` is not None. We should evaluate all these Optionals and see if they should ever be None.

I didn't want to pollute the code with `if None... raise` checks all over the place.

cihati (approved):

PR #36, opened 2022-05-23
Relax rich contraint

dpp23:
https://github.com/dbt-labs/dbt-core has a dependency constraint `typing-extensions (>=3.7.4,<3.11)` which conflicts with `rich==12` which has `typing-extensions = { version = ">=4.0.0, <5.0", python = "<3.9" }`

Relax `rich` constraint to allow the use of `layer` with `dbt-core`


eminn (approved):
Assuming that it won't break and console interaction.

dpp23:
> Would you also update the PR description as two why you need this? There was a breaking on version 11.0 (the one about RenderGroup, https://github.com/Textualize/rich/releases?page=1 ); so if we do use 10.x then we'll have an error

Thanks @cihati I've updated the PR to ensure we have `rich>=11` and also updated the PR description with the reasoning behind the change.

PR #37, opened 2022-05-23
Ensure removed dependencies get cleaned up from poetry vevn

dpp23:
By default poetry won't uninstall a package if it is completely removed as a dependency. Because of this, I almost deleted a required dependency in https://github.com/layerai/sdk/pull/35

Force poetry to remove packages and only use dependencies in the lock file.


cihati (approved):
I think I got bitten by this too!

PR #39, opened 2022-05-23
Add numpy ndarray serialisation to arrow

donatasm:
Add multidimensional `numpy` array support in layer datasets.

Arrays must be wrapped with the `layer.Arrays` class to provide the conversions between the `numpy.ndarray ` and `arrow` types. Numpy arrays are serialised to `pyarrow.binary` type, using [numpy.save](https://numpy.org/doc/stable/reference/generated/numpy.save.html) and deserialised using [numpy.load](https://numpy.org/doc/stable/reference/generated/numpy.load.html).

```python
import pandas as pd
import numpy as np

import layer
from layer.decorators import dataset

@dataset("ndarrays_demo")
def build():
    return pd.DataFrame(
        {
            "array": layer.Arrays(
                (
                    np.array([[1, 2, 3], [4, 5, 6]], dtype=np.int64),
                    np.array([[7, 8, 9], [10, 11, 12]], dtype=np.int64),
                    np.array([[13, 14, 15], [16, 17, 18]], dtype=np.int64),
                )
            )
        },
        copy=False,
    )
```

### DIFF
--- a/layer/pandas_extensions.py
+++ b/layer/pandas_extensions.py
@@ -170,8 +170,137 @@ def _image_bytes(image: Image) -> bytes:
         return buf.getvalue()
 
 
+_TYPE_NDARRAY_NAME = "layer.ndarray"
+
+
+class _ArrayType(pa.ExtensionType):
+    """Arrow type extension https://arrow.apache.org/docs/python/extending_types.html#defining-extension-types-user-defined-types
+    Provides addtional metadata for arrow schema about the ndarray format.
+    """
+
+    def __init__(self) -> None:
+        pa.ExtensionType.__init__(self, pa.binary(), _TYPE_NDARRAY_NAME)
+
+    def __arrow_ext_serialize__(self) -> bytes:
+        return b""
+
+    @classmethod
+    def __arrow_ext_deserialize__(
+        cls, storage_type: pa.DataType, serialized: bytes
+    ) -> pa.ExtensionType:
+        return _ArrayType()
+
+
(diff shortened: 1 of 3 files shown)


layer/pandas_extensions.py
```diff
@@ -170,8 +170,137 @@ def _image_bytes(image: Image) -> bytes:
...
+        total_bytes = 0
+        for arr in self._arrays:
+            total_bytes += arr.nbytes
+        return total_bytes
+
+
+def _array_bytes(arr: np.ndarray) -> bytes:  # type: ignore
+    with io.BytesIO() as buf:
+        np.save(buf, arr, allow_pickle=False, fix_imports=False)  # type: ignore
+        return buf.getvalue()
```
cihati, line 296: nit: why is this not a method of `class Arrays`. same for `_ArrayType` and `_load_array`

cihati (approved):
Presumably, we'll update the docs for the newly exposed `Images` and `Arrays` in a follow-up PR?

donatasm:
> Presumably, we'll update the docs for the newly exposed `Images` and `Arrays` in a follow-up PR?

Yes, the docs are incoming.

layer/pandas_extensions.py
donatasm: > nit: why is this not a method of `class Arrays`. same for `_ArrayType` and `_load_array`

As `self` is not being accessed, I prefer keeping them as module functions.

PR #41, opened 2022-05-24
Added project url output

mecevit:


layer/contracts/projects.py
```diff
@@ -132,6 +132,18 @@ def with_readme(self, readme: Optional[str]) -> "Project":
     def with_functions(self, functions: Sequence[Function]) -> "Project":
         return replace(self, functions=functions)
 
+    def __str__(self) -> str:
+        from layer.config import DEFAULT_PATH, ConfigManager
+
```
eminn, line 137: maybe we can read the config only if there is an account ?

mecevit: Done

eminn:
I think you need to rebase or merge master

layer/contracts/projects.py
```diff
@@ -132,6 +132,18 @@ def with_readme(self, readme: Optional[str]) -> "Project":
...
+        if self.account:
+            from layer.config import DEFAULT_PATH, ConfigManager
+
+            config = ConfigManager(DEFAULT_PATH).load()
+            return f"Your Layer project is here: {config.url}/{self.account.name}/{self.name}"
+        else:
+            return f"Project({self.name})"
+
+    def __repr__(self) -> str:
+        return self.__str__()
```
cihati, line 145: So, this is not technically correct & intended use of `__repr__`. Basically, this should return a string that when used inside `eval()`, actually creates the object. For example `eval(repr(some_project_object))` should be equal to `some_project_object`. See point 4 in https://betterprogramming.pub/advanced-python-9-best-practices-to-apply-when-you-define-classes-871a27af658b for a better example. Either implement `__repr__` accordingly, or better not touch it.

mecevit: The thing is Jupyter and Google Colab uses `__repr__` to print the string representation of the object. Should we support users to create an instance of Layer Project with a string representation?

cihati: Talked offline with Mehmet, let's just merge this as is, I'll take a look at this in the future.

PR #44, opened 2022-05-25
Remove pillow dependency from SDK

donatasm:
Remove pillow dependency from SDK, to avoid unnecessary version constraint conflicts. Only instance checks are made on `PIL`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ POETRY := $(shell command -v poetry 2> /dev/null)
 install: $(INSTALL_STAMP) ## Install dependencies
 $(INSTALL_STAMP): pyproject.toml poetry.lock
 	@if [ -z $(POETRY) ]; then echo "Poetry could not be found. See https://python-poetry.org/docs/"; exit 2; fi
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -9,7 +9,7 @@ POETRY := $(shell command -v poetry 2> /dev/null)
 install: $(INSTALL_STAMP) ## Install dependencies
 $(INSTALL_STAMP): pyproject.toml poetry.lock
 	@if [ -z $(POETRY) ]; then echo "Poetry could not be found. See https://python-poetry.org/docs/"; exit 2; fi
-	$(POETRY) install --remove-untracked
+	$(POETRY) install
```
dpp23, line 12: Can we keep this one as it was for now? I will work on fixing it today, but if we remove `--remove-untracked` Pillow will remain in all the venvs that are used in CI. 

I almost removed a dependency that was actually needed [here](https://github.com/layerai/sdk/pull/35) because of it.

dpp23 (approved):

PR #55, opened 2022-05-26
Add model prediction API

aleksmitov:
Introduce a `Model.predict(input_df: pandas.Dataframe) -> pandas.Dataframe` API.
Implemented for sci-kit learn and catboost models.

### DIFF
--- a/layer/flavors/base.py
+++ b/layer/flavors/base.py
@@ -1,7 +1,9 @@
 import inspect
 from abc import ABCMeta, abstractmethod, abstractproperty
 from pathlib import Path
+from typing import Callable, Tuple
 
+import pandas as pd
 from layerapi.api.entity.model_version_pb2 import ModelVersion
 
 from layer.types import ModelArtifact
@@ -64,7 +66,9 @@ class ModelFlavor(metaclass=ABCMeta):
         """
 
     @abstractmethod
-    def load_model_from_directory(self, directory: Path) -> ModelArtifact:
+    def load_model_from_directory(
+        self, directory: Path
+    ) -> Tuple[ModelArtifact, Callable[[pd.DataFrame], pd.DataFrame]]:
         """Defines the method that this Model Flavor uses to load a model from a directory.
 
         Returns:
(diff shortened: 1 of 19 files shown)


layer/flavors/base.py
```diff
@@ -1,12 +1,21 @@
...
 
+import pandas as pd
 from layerapi.api.entity.model_version_pb2 import ModelVersion
 
 from layer.types import ModelArtifact
 
 
+@dataclass(frozen=True)
+class ModelRuntimeObjects:
+    model_artifact: ModelArtifact
```
eminn, line 15: maybe rename to model_object ? and rename the type as well

layer/flavors/catboost.py
```diff
@@ -20,7 +22,15 @@ def save_model_to_directory(
...
         import mlflow.catboost
 
-        return mlflow.catboost.load_model(directory.as_uri())
+        model = mlflow.catboost.load_model(directory.as_uri())
+        return ModelRuntimeObjects(
+            model, lambda input_df: self.__predict(model, input_df)
+        )
+
+    @staticmethod
+    def __predict(model: Any, input_df: pd.DataFrame) -> pd.DataFrame:
```
eminn, line 34: why do we have underscores we do expect our plugins to call this method.

aleksmitov: The plugin will call Model.predict(), i.e the object we return from layer.get_model(), so it's not going to be this one.

aleksmitov: I'm talking about `layer/contracts/models.py.`predict() in this PR.

layer/flavors/base.py
aleksmitov: I've renamed it.

layer/flavors/catboost.py
eminn: I see.  Further comments below:
- We can also change model parameter type to `ModelObject `
- Do we need this method to be static ?

aleksmitov: - Updated the code based on your first point.
- About the second, I think it should remain a static method. I don't see why it should be reading/writing state from the model flavor class instance. The other option is changing the model runtime object that has been loaded with mlflow. I don't think we should do that either.

eminn (approved):

PR #62, opened 2022-05-27
Ignore all tests segfaulting on Mac M1

donatasm:
Ignore all tests segfaulting on Mac M1.

### DIFF
--- a/test/unit/test_model_flavor.py
+++ b/test/unit/test_model_flavor.py
@@ -21,8 +21,8 @@ from layer.flavors.utils import get_flavor_for_model
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


dpp23 (approved):

test/unit/test_model_flavor.py
```diff
@@ -21,8 +21,8 @@
 logger = logging.getLogger(__name__)
 
 
+@pytest.mark.skipif(platform.system() == "Darwin", reason="Segfaults on Mac")
```
dpp23, line 24: Maybe we can change this to `platform.processor() == 'arm'`?

donatasm: The check was there previously on `Darwin', so it did not work for someone else on Mac as stated in the reason. `arm` would turn them of for Linux as well.

PR #63, opened 2022-05-27
Added `Custom Model` support

mecevit:
Added Custom Model support to provide a higher level API for users to:
- Develop their own custom model flavor
- Create a high level inference api for complex prediction pipelines

## How to use
```python
class SentimentAnalysisModel(layer.CustomModel):
    def __init__(self):
        self.model = layer.get_model("...")
        self.tokenizer = layer.get_model("...")
        
    def predict(self, input):
        output = self.model(self.tokenizer(input))
        return output

@model("sentimentanalysis")
def train():
    return SentimentAnalysisModel()

layer.run([train])
```

Later you use the registered custom model with:
```python
custom_model = layer.get_model("sentimentanalysis")
custom_model("I really hated this movie!")

# > Negative
```


eminn (approved):

layer/flavors/custom.py
```diff
@@ -0,0 +1,51 @@
...
+    A generic model that evaluates inputs and produces outputs.
+    """
+
+    def __init__(self) -> None:
+        """
+        Initializes the custom  model
+        """
+
+    @abstractmethod
+    def predict(self, model_input: Any) -> Any:
```
eminn, line 24: Are we going to accept any input ? Shouldn't it be a dataframe ?

aleksmitov, line 24: To provide support for the new prediction api for such custom models, this function should have more specific signature. E.g it accepts a pandas dataframe and returns a pandas dataframe

mecevit: Fixed

mecevit: Fixed

PR #64, opened 2022-05-27
Implicitly convert images and ndarrays

donatasm:
With this change, there is no need to wrap pandas data frame image or ndarray columns into `layer.Images` or `layer.Arrays`. Instead, `store_dataset` will figure this out by checking if the instances are `PIL.Image.Image` and `numpy.ndarray` and do the conversion automatically.

### DIFF
--- a/layer/pandas_extensions.py
+++ b/layer/pandas_extensions.py
@@ -346,3 +346,21 @@ def _register_type_extensions() -> None:
     # register pandas extension types
     register_extension_dtype(_ImageDtype)
     register_extension_dtype(_ArrayDtype)
+
+
+def _infer_custom_types(data: pd.DataFrame) -> pd.DataFrame:
+    if len(data) == 0:
+        return data
+    first_row = data.head(n=1)
+    try:
+        from PIL.Image import Image
+
+        for col in first_row:
+            value = data[col][0]
+            if isinstance(value, Image):
+                data[col] = Images(data[col])
+            elif isinstance(value, np.ndarray) and value.ndim > 1:
+                data[col] = Arrays(data[col])
+    except ImportError:
+        pass
+    return data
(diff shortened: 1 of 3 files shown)


layer/pandas_extensions.py
```diff
@@ -346,3 +346,21 @@ def _register_type_extensions() -> None:
     # register pandas extension types
     register_extension_dtype(_ImageDtype)
     register_extension_dtype(_ArrayDtype)
+
+
+def _infer_custom_types(data: pd.DataFrame) -> pd.DataFrame:
```
cihati, line 351: nit: why the prefix `_`?

test/unit/test_pandas_extensions.py
```diff
@@ -268,3 +277,49 @@ def test_pandas_arrays_describe():
...
+    inferred = _infer_custom_types(data)
+    assert_frame_equal(data, inferred)
+
+
+def test_infer_custom_types_infers_images_type(tmp_path):
+    data = pd.DataFrame({"col1": [1, 2, 3], "img": [_image(1), _image(2), _image(3)]})
+    inferred = _infer_custom_types(data)
+    parquet_path = tmp_path / str(uuid4())
+    data.to_parquet(parquet_path, engine=_PARQUET_ENGINE)
+    data_parquet = pd.read_parquet(parquet_path, engine=_PARQUET_ENGINE)
```
cihati, line 298: why does *this* unit test have this logic? shouldn't we only test `infer_custom_types`? (same for other tests below)

especially given we already have `test_pandas_images_read_write` and `test_pandas_arrays_read_write`.

cihati (approved):
approving with a comment

layer/pandas_extensions.py
donatasm: It is internal function, should not be used outside `layer`.

cihati: It wouldn't be? as we don't export it in https://github.com/layerai/sdk/blob/main/layer/__init__.py

test/unit/test_pandas_extensions.py
donatasm: To make sure data frame is still written and read correctly after the implicit type conversion. I put a comment for duplication

layer/pandas_extensions.py
donatasm: Yes, but is still private, and `_` emphasises it should stay like that.

PR #71, opened 2022-05-30
Skip segfaulting tests only on Mac M1

donatasm:
Skip segfaulting tests only on Mac M1. Related to suggestion in https://github.com/layerai/sdk/pull/62#discussion_r883413545.

Doing a substring check, as other methods still return i386 when run through `Rosetta` emulation.

### DIFF
--- a/test/unit/test_train.py
+++ b/test/unit/test_train.py
@@ -11,7 +11,7 @@ from layer.config import ClientConfig
 from layer.exceptions.exceptions import UnexpectedModelTypeException
 from layer.training.train import Train
 
-from .. import IS_DARWIN_ARM64
+from .. import IS_DARWIN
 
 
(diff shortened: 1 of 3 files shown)


dpp23 (approved):

test/unit/test_train.py
```diff
@@ -108,7 +109,7 @@ def test_train_raises_exception_if_error_happens() -> None:
         set(),
     ],
 )
-@pytest.mark.skipif(platform.system() == "Darwin", reason="Segfaults on Mac")
+@pytest.mark.skipif(IS_DARWIN_ARM64, reason="Segfaults on Mac M1")
```
eminn, line 112: This was failing on regular mac os as well with a segfault.

PR #77, opened 2022-05-31
Support more model flavors with the prediction API

aleksmitov:


layer/flavors/tensorflow.py
```diff
@@ -35,9 +38,16 @@ def load_model_from_directory(self, directory: Path) -> ModelRuntimeObjects:
...
-            model, lambda input_df: self.__predict(model, input_df)
+            model, lambda input_df: self.__predict(directory, input_df)
         )
 
     @staticmethod
-    def __predict(model: ModelObject, input_df: pd.DataFrame) -> pd.DataFrame:
-        raise Exception("Not implemented")
+    def __predict(directory: Path, input_df: pd.DataFrame) -> pd.DataFrame:
+        cache_key = str(directory)
+        if cache_key not in LAYER_CACHED_TF_MODELS:
```
eminn, line 47: Why do we implement caching for only tf models ?

eminn (approved):

mecevit (approved):

layer/flavors/tensorflow.py
aleksmitov: 1) We still want to load tf models with mlflow.tensorflow.load_model() In order to provide users with the full tf model API.
2) Unfortunately, there's no straightforward way to convert a tf model to the underling mlflow model wrapper that exposes the prediction api from/to pandas dataframes.
3) As a result, for the pd.Dataframe -> pd.Dataframe function, we cannot reuse the object loaded with mlflow.tensorflow.load_model(), but we need to load it with mlflow.pyfunc.load_model(). And we finally apply caching to it, to avoid reloading the tf pyfunc model on each .predict() call.

For the other models this is not necessary as either their native model flavors already support a pd.Dataframe -> pd.Dataframe API or their have an easy to use mlflow model wrappers that we can use.

PR #83, opened 2022-06-01
Fix custom model pickling

mecevit:
Previous custom model pickle routine was using Cloudpickle which was causing errors to load a custom model in different python versions. Now we use plain pickle to serialize and load the model.

### DIFF
--- a/layer/flavors/custom.py
+++ b/layer/flavors/custom.py
@@ -105,7 +105,9 @@ class CustomModelFlavor(ModelFlavor):
             spec.loader.exec_module(module)  # type: ignore
             sys.modules[model_module_name] = module
         except Exception:
-            raise Exception("Failed to load model. Please clear your cache with `layer.clear_cache` and try again!")
+            raise Exception(
+                "Failed to load model. Please clear your cache with `layer.clear_cache` and try again!"
+            )
 
         # Load model itself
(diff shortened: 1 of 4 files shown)


layer/flavors/custom.py
```diff
@@ -44,12 +46,43 @@ def save_model_to_directory(
...
+        # Store class config
+        with open(directory / self.MODEL_CONFIG_FILE, mode="w") as file:
+            custom_model_class = type(model_object)
+            class_config = custom_model_class.__module__ + "." + custom_model_class.__name__
+            file.write(class_config)
+
+        # Store class definition
+        import inspect
+        custom_class_model = type(model_object)
+        class_model_source_code = inspect.getsource(custom_class_model)
```
eminn, line 59: Can we add a check to make sure we are getting the source or throw an error so that user needs to define it in another file ? How about notebooks will it work on that case ?

mecevit: Ok, adding checks. 

Yes it works in the notebook case with the only limitation to implement the custom model in a seperate python file.

eminn (commented):
should we document the new requirement of saving the custom model into its own file ?

mecevit:
> should we document the new requirement of saving the custom model into its own file ?

Yes, i will write a detailed documentation in our docs in a follow up PR

eminn (approved):

PR #85, opened 2022-06-01
Add Keras and XGBoost callbacks

cihati:
Implements convenience classes/methods for Keras and XGBoost.


layer/logged_data/callbacks.py
```diff
@@ -0,0 +1,59 @@
...
+        df = pd.DataFrame(data=importance, columns=["feature", "importance"])
+        layer.log({"Feature Importance Table": df})
+        layer.log(
+            {"Feature Importance": df.plot.bar(x="feature", y="importance", rot=0)}  # type: ignore
+        )
+
+        return model
+
+    def after_iteration(self, model: Any, epoch: int, evals_log: dict) -> bool:  # type: ignore
+        layer.log(evals_log, epoch)
```
eminn, line 46: similarly here

eminn (commented):
Is it possible we could write some unit tests on a basic train so that we can verify the calls to layer.log after registering the callbacks ?

eminn (approved):
💯

PR #92, opened 2022-06-03
Support dict for layer.log

cihati:
This make it slightly convenient to log tabular data, as an alternative to `pandas.DataFrame`.


mecevit (approved):
Thank you!

PR #95, opened 2022-06-03
Yuranos/add e2e tests for layerlog

yuranos:
Add e2e tests for `layer.log` functions for all supported log types.


test/e2e/test_log_types.py
```diff
@@ -0,0 +1,159 @@
...
+from layer.contracts.projects import Project
+from layer.decorators import dataset
+from test.e2e.assertion_utils import E2ETestAsserter
+
+
+def test_scalar_values_logged(
+    initialized_project: Project, asserter: E2ETestAsserter, client: LayerClient
+):
+    # when
+    layer.init(initialized_project.name)
```
donatasm, line 15: Isn't the project `initialized_project` (as the name implies) already initialised?

test/e2e/test_log_types.py
```diff
@@ -0,0 +1,159 @@
...
+    logged_data = client.logged_data_service_client.get_logged_data(
+        tag="int-log", dataset_build_id=first_ds.build.id
+    )
+    assert logged_data.data == "123"
+    assert logged_data.logged_data_type == LoggedDataType.NUMBER
+    assert logged_data.tag == "int-log"
+
+
+def test_pandas_dataframe_logged(initialized_project: Project):
+    # when
```
donatasm, line 58: What is the point of having such kind of comments?

test/e2e/test_log_types.py
```diff
@@ -0,0 +1,159 @@
...
+
+    # given
+    @dataset(dataset_name)
+    def scalar():
+        data = [[1, "product1", 15], [2, "product2", 20], [3, "product3", 10]]
+        dataframe = pd.DataFrame(data, columns=["Id", "Product", "Price"])
+        layer.log(
+            {
+                "str-log": "bar",
+                "int-log": 123,
```
cihati, line 27: add bool and float too

cihati (commented):
* Can you please add video & markdown & dictionary tests too?
* (arguable) In this case, might be worth having a single test and multiple layer.log statements; we would avoid quite a bit of overhead of creating new datasets and could speed things up quite a bit IMO.

test/e2e/test_log_types.py
yuranos: I actually had them, and even scientific notation. Just forgot to Put them back after merging with the existing test that I initially didn't notice and that was written by someone else. Will do!

test/e2e/test_log_types.py
yuranos: Thanks, removed.

test/e2e/test_log_types.py
yuranos: Done.

yuranos:
> * Can you please add video & markdown & dictionary tests too?
> * (arguable) In this case, might be worth having a single test and multiple layer.log statements; we would avoid quite a bit of overhead of creating new datasets and could speed things up quite a bit IMO.

Added missing types. Also, merged some tests together, but not all of them. At this points it doesn't seem like a big overhead, but improved debugging when tests can be run separately(also arguably, of course).

test/e2e/test_log_types.py
```diff
@@ -0,0 +1,244 @@
...
+    dataset_func()
+
+    ds = client.data_catalog.get_dataset_by_name(initialized_project.id, ds_name)
+
+    logged_data = client.logged_data_service_client.get_logged_data(
+        tag=ds_tag, dataset_build_id=ds.build.id
+    )
+
+    print("logged_data:")
+    print(logged_data)
```
cihati, line 117: ?

test/e2e/test_log_types.py
```diff
@@ -0,0 +1,244 @@
...
+
+def test_matplotlib_objects_logged(initialized_project: Project, client: LayerClient):
+    # given
+    figure_tag = "matplotlib_figure_tag"
+    plot_tag = "matplotlib_pyplot_tag"
+
+    ds_name = "ds_with_plots"
+
+    @dataset(ds_name)
+    # @pip_requirements(packages=["matplotlib==3.5.1"])
```
cihati, line 181: ?

cihati (approved):

test/e2e/test_log_types.py
yuranos: Oh, sorry:)

test/e2e/test_log_types.py
yuranos: Done.

PR #106, opened 2022-06-07
Add e2e test for pandas datetime[ns] type

donatasm:


cihati (approved):
I'd move this above near other date/time checks. also, do we need for ms as well?

donatasm:
> I'd move this above near other date/time checks. also, do we need for ms as well?

Pandas `Timestamp` type is [always mapped to nano resolution in pyarrow](https://arrow.apache.org/docs/python/pandas.html#pandas-arrow-conversion), so no need to support other precisions for now.

PR #107, opened 2022-06-07
Use project full name as much as possible

adrien-layer:
We introduce `ProjectFullName` and promote `account-name/project-name` as project identifier as much as possible without changing existing APIs.

When `account_id` or `project_id` are needed we limit their existence to the grpc clients.
Later, we'd like the APIs to take the project path.

We also remove some Optionals to make the code more understandable, less complex and less defensive

Finally, we fix an issue with how functions' explicit dependencies paths were made absolute.


volkangurel (approved):

layer/global_context.py
```diff
@@ -45,6 +48,14 @@ def current_project_name() -> Optional[str]:
     return _GLOBAL_CONTEXT.project_name
 
 
+def set_current_account_name(name: Optional[str]) -> None:
```
marta-tatiana, line 51: I wonder if we need to expose `set_current_account_name` and `set_current_project_name` separately? 
it feels like an opportunity to make a mistake and get into some mismatch. Could we go with `set_current_project_full_path` or is that impossible at this moment?

PR #109, opened 2022-06-09
Simplify setup pre-requirements on Apple arm processors

almg80:
Avoid requiring specific package versions added manually to Conda environment and detect correct environment is available

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,12 +4,36 @@ E2E_TEST_HOME := $(ROOT_DIR)/build/e2e-home
 TEST_TOKEN_FILE := .test-token
 POETRY := $(shell command -v poetry 2> /dev/null)
 IN_VENV := $(shell echo $(CONDA_DEFAULT_ENV)$(CONDA_PREFIX)$(VIRTUAL_ENV))
+CONDA_ENV_NAME := $(shell echo $(CONDA_DEFAULT_ENV))
+UNAME_SYS := $(shell uname -s)
+UNAME_ARCH := $(shell uname -m)
 
 .DEFAULT_GOAL:=help
 
 install: $(INSTALL_STAMP) ## Install dependencies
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -4,12 +4,36 @@ E2E_TEST_HOME := $(ROOT_DIR)/build/e2e-home
...
+	$(eval LIGHTGBM_VERSION := $(shell poetry show lightgbm --no-ansi --no-dev | grep version | awk '{print $$3}'))
+	$(eval H5PY_VERSION := $(shell poetry show h5py --no-ansi --no-dev | grep version | awk '{print $$3}'))
+	$(eval PYARROW_VERSION := $(shell poetry show pyarrow --no-ansi --no-dev | grep version | awk '{print $$3}'))
+	@conda install -y tokenizers==$(TOKENIZERS_VERSION) xgboost==$(XGBOOST_VERSION) lightgbm==$(LIGHTGBM_VERSION) h5py==$(H5PY_VERSION) pyarrow==$(PYARROW_VERSION)
+endif
+else
+	@echo 'Not inside a conda environment or conda not installed, this is a requirement for Apple arm processors'
+	@echo 'See https://github.com/conda-forge/miniforge/'
+	@exit 1
+endif
```
dpp23, line 34: Can we move this to a function?

Maybe we can also move the repetetion of `$(shell poetry show tokenizers --no-ansi --no-dev | grep version | awk '{print $$3}'))` to a function, but that is a lesser concern if it is all outside of the `install` step.

almg80: @dpp23 Can you review the latest commit? Split `get_python_package_version` into a function but the remaining code gets too messy in a function so moved it instead into a target and made it a pre-requirement for `install` wdyt?

dpp23 (approved):

PR #110, opened 2022-06-09
Remove all references to model train params

yuranos:
## Description
As we are redesigning the way we add logs for our entities, model train parameters endpoints will no longer be used. They are also removed from Layer backend.


yuranos:
> __parse_parameter

@cihati , done.

cihati (approved):

PR #122, opened 2022-06-16
Upgrade to v0 9 365026

adrien-layer:
More APIs support passing the project full name now so we can make the account explicit in more places

### DIFF
--- a/layer/main.py
+++ b/layer/main.py
@@ -41,6 +41,7 @@ from layer.exceptions.exceptions import (
 )
 from layer.global_context import (
     current_account_name,
+    current_project_full_name,
     current_project_name,
     get_active_context,
     reset_active_context,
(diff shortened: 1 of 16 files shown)


layer/main.py
```diff
@@ -468,17 +469,20 @@ def start_train(
         train.save_model(trained_model)
 
     """
-    project_name = current_project_name()
-    if not project_name:
-        raise Exception("Missing current project name")
+    project_full_name = current_project_full_name()
+    if not project_full_name:
+        raise Exception(
+            "Missing current project name, please use 'set_current_project_full_name(\"account-name/project-name\")'"
```
marta-tatiana, line 475: nit: I think we should only suggest using layer.init(). 
feel free to skip this comment if this PR needs to be merged to unblock things.

adrien-layer: good shout

adrien-layer: fixed

PR #124, opened 2022-06-16
Compute number of rows per batch slice dynamically

donatasm:
Instead of having a hardcoded chunk size set to a very small value, compute the chunk size of the batch dynamically. This should significantly improve the throughput for `store_dataset`.

### DIFF
--- a/layer/clients/data_catalog.py
+++ b/layer/clients/data_catalog.py
@@ -1,10 +1,9 @@
-import math
 import tempfile
 import uuid
 from contextlib import contextmanager
 from logging import Logger
 from pathlib import Path
-from typing import Any, Callable, Iterator, List, Optional, Sequence
+from typing import Any, Callable, Generator, Iterator, List, Optional, Sequence
 
 import pandas
 import pyarrow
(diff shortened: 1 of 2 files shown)


layer/clients/data_catalog.py
```diff
@@ -439,3 +427,34 @@ def _list_datasets(self, sort_fields: Sequence[SortField] = ()) -> List[PBDatase
                 ),
             ).datasets
         )
+
+
+def _get_batch_chunks(
+    batch: pyarrow.RecordBatch, max_chunk_size_bytes: int = 4_000_000
+) -> Generator[pyarrow.RecordBatch, None, None]:
+    """
+    Slice the batch into the chunks, based on average row size,
```
maasg, line 436:
```suggestion
    Slice the batch into chunks, based on average row size,
```

layer/clients/data_catalog.py
```diff
@@ -439,3 +427,34 @@ def _list_datasets(self, sort_fields: Sequence[SortField] = ()) -> List[PBDatase
...
+    but not exceeding the maximum chunk size.
+    """
+
+    bytes_per_row = batch.nbytes / batch.num_rows  # average row size
+    max_num_rows = int(max_chunk_size_bytes / bytes_per_row)  # rows per chunk
+    if max_num_rows == 0:
```
maasg, line 442: It would be easier to read if we compare `if bytes_per_row > max_chunk_size_bytes`  instead of the result of the integer division.

maasg (approved):
Looks good. I left a non-blocking suggestion for your consideration.

PR #139, opened 2022-06-20
Check supported Python version

cihati:
Check if the Python version is supported on the first remote call (when users call `layer.run`) since fetching models/dataset or running local builds/trains works in any Python version.

### DIFF
--- a/layer/main.py
+++ b/layer/main.py
@@ -44,10 +44,12 @@ from layer.global_context import (
     current_project_full_name,
     current_project_name,
     get_active_context,
+    has_shown_python_version_message,
     has_shown_update_message,
     reset_active_context,
     reset_to,
     set_active_context,
+    set_has_shown_python_version_message,
     set_has_shown_update_message,
 )
 from layer.logged_data.log_data_runner import LogDataRunner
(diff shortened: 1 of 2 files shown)


layer/main.py
```diff
@@ -665,11 +668,26 @@ def _check_latest_version() -> None:
     current_version = pkg_resources.get_distribution("layer").version
     if current_version != latest_version:
         print(
-            "You are using the version {current_version} but the latest version is {latest_version}, please upgrade with 'pip install --upgrade layer'"
+            f"You are using the version {current_version} but the latest version is {latest_version}, please upgrade with 'pip install --upgrade layer'"
```
cihati, line 671: I forgot this in the previous PR

layer/main.py
```diff
@@ -665,11 +668,26 @@ def _check_latest_version() -> None:
...
 
+def _check_python_version() -> None:
+    if has_shown_python_version_message():
+        return
+
+    import platform
+
+    major, minor, _ = platform.python_version_tuple()
+
+    if major != "3" or minor not in ["7", "8"]:
```
eminn, line 684: What about minor 9 and 10 ?

eminn (approved):

layer/main.py
cihati: I think we only support Python 3.7 and 3.8? cc @mecevit

mecevit: @eminn you should know this better, because of cloudpickle limitations, i believe we only 3.7 + 3.8 images for our executors, no?

aleksmitov (approved):

PR #140, opened 2022-06-20
Support list of dict-s as output of a dataset build function

aleksmitov:
To support https://www.notion.so/layerco/SDK-Devex-v1-0ee60206f0df4b70a98e379a60bc45ba#d2049beb08e940b98cf2cda25ae70f3a

After this gets merged, I'll follow up with a PR that uses this logic on the backend executors side.

### DIFF
--- a/layer/utils/runtime_utils.py
+++ b/layer/utils/runtime_utils.py
@@ -1,0 +1,23 @@
+from typing import Any
+
+import pandas as pd
+
+
+def check_and_convert_to_df(result: Any) -> pd.DataFrame:
+    allowed_types = [pd.DataFrame, list]
+    if not any([isinstance(result, x_type) for x_type in allowed_types]):
+        raise Exception(
+            f"Unsupported return type. Expected one of {allowed_types}, got: {type(result)}"
+        )
+    if isinstance(result, pd.DataFrame):
+        if len(result.columns) == 0:
+            raise Exception("pandas.DataFrame object has 0 columns")
+        return result
+    if isinstance(result, list):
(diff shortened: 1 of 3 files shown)


layer/utils/runtime_utils.py
```diff
@@ -0,0 +1,23 @@
...
+            raise Exception("pandas.DataFrame object has 0 columns")
+        return result
+    if isinstance(result, list):
+        dict_elements = all([isinstance(list_element, dict) for list_element in result])
+        if not dict_elements:
+            raise Exception("Expected a list of dict elements")
+        return pd.DataFrame.from_records(result)  # type: ignore
+    raise Exception(
+        f"Unsupported return type. Expected one of {allowed_types}, got: {type(result)}"
+    )
```
cihati, line 23: why is this needed?

layer/utils/runtime_utils.py
```diff
@@ -0,0 +1,23 @@
+from typing import Any
```
cihati, line 1: Will this file be used in the executors? If not, perhaps I'd move it under `layer/decorators/` folder.

cihati (approved):

cihati (commented):
Also, please update the docs of dataset_decorator.py

layer/utils/runtime_utils.py
aleksmitov: Linter complains otherwise

layer/utils/runtime_utils.py
aleksmitov: The intention is to be used by an executor, yes.

PR #127, opened 2022-06-17
Re-organize decorator files

volkangurel:
Re-organize decorator files to improve imports and make it possible to import decorators from the top level `layer` module.

Fixes LAY-2941.

PR #143, opened 2022-06-20
Consolidate function definitions

volkangurel:
Consolidate function definitions

This PR is one of several to split https://github.com/layerai/sdk/pull/127 up. This PR should merge after #137 is merged.

### DIFF
--- a/layer/contracts/runs.py
+++ b/layer/contracts/runs.py
@@ -1,4 +1,3 @@
-import abc
 import hashlib
 import inspect
 import os
@@ -15,7 +14,7 @@ from layerapi.api.entity.run_pb2 import Run as PBRun
 from layerapi.api.ids_pb2 import RunId
 
 from layer.config import DEFAULT_FUNC_PATH
-from layer.exceptions.exceptions import LayerClientException
+from layer.contracts.assertions import Assertion
 
 from .asset import AssetPath, AssetType
 from .fabrics import Fabric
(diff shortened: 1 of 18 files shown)


layer/decorators/layer_wrapper.py
```diff
@@ -62,3 +65,44 @@ def __init__(
...
+                if d.project_name is not None:
+                    full_path_dep = d.with_project_full_name(
+                        ProjectFullName(
+                            account_name=current_project_full_name.account_name,
+                            project_name=d.project_name,
+                        )
+                    )
+                else:
+                    full_path_dep = d.with_project_full_name(current_project_full_name)
+            asset_dependencies.append(full_path_dep)
```
cihati, line 87: nit: I'd put this into a separate method. same for get_pip_dependencies below

test/unit/projects/test_project_hash_calculator.py
```diff
@@ -55,3 +59,20 @@ def func3():
 
     # then
     assert first_hash != second_hash
+
+
+def _make_test_function_definition(
```
cihati, line 64: https://github.com/layerai/sdk/pull/127#discussion_r901118767

layer/contracts/runs.py
```diff
@@ -198,53 +176,8 @@ def _pack(self) -> None:
...
             cloudpickle.dump(self.func, file, protocol=pickle.DEFAULT_PROTOCOL)
 
-        # Add requirements to tarball
-        if self.pip_requirements_file:
-            shutil.copy(self.pip_requirements_file, self.environment_path)
-        elif self.pip_packages:
-            with open(self.environment_path, "w") as reqs_file:
-                reqs_file.writelines(
-                    list(map(lambda package: f"{package}\n", self.pip_packages))
-                )
```
cihati, line 208: you seem to be changing this part as part of this refactor -- did you test this thoroughly?

layer/contracts/runs.py
```diff
@@ -61,41 +60,41 @@ def local_relative_paths(self) -> Iterator[str]:
...
-        self.resource_paths = layer_settings.get_resource_paths()
-        fabric = layer_settings.get_fabric()
-        if fabric is None:
-            fabric = Fabric.default()
-        self._fabric = fabric
-        self._dependencies: List[AssetPath] = layer_settings.get_dependencies()
-        self.pip_packages = layer_settings.get_pip_packages()
-        self.pip_requirements_file = layer_settings.get_pip_requirements_file()
 
         self.func_source = inspect.getsource(self.func)
```
cihati, line 100: so, the `source_code_digest`, also `uri` above etc are all dataset-specific. I guess one downside of this "bundling" is that models will have these unnecessarily as well..

cihati (approved):

cihati:
@volkangurel I think this got automerged, but please take a note of my comments = )

layer/decorators/layer_wrapper.py
volkangurel: done

test/unit/projects/test_project_hash_calculator.py
volkangurel: done

layer/contracts/runs.py
volkangurel: yes, true, but with several more factors we should be able to get datasets and models closer and more consistent with each other. for example, no reason why `source_code_digest` and `uri` can't be attributes of both datasets and models.

layer/contracts/runs.py
volkangurel: not sure what is thoroughly, but tested with several remote execution examples that I think should cover this.

PR #145, opened 2022-06-21
Log system metrics

micheldlebeau:
This PR logs system metrics for remote and local model trains.
One caveat for now, if several GPUs are on a machine, all GPU stats that Nvidia SMI can access will be shown. This might not be the desired behaviour when we start having remote trainings with several shared GPUs.

### DIFF
--- a/layer/training/runtime/model_trainer.py
+++ b/layer/training/runtime/model_trainer.py
@@ -384,7 +384,6 @@ class ModelTrainer:
                     ModelTrainStatus.TRAIN_STATUS_IN_PROGRESS,
                     self.logger,
                 )
-                sleep(1)  # temporarily make things slower
                 self.logger.info("Executing the train_model_func")
                 work_dir = self.train_context.get_working_directory()
                 os.chdir(work_dir)
@@ -394,13 +393,10 @@ class ModelTrainer:
                     train_model_func.__name__,
                     target_dir=str(work_dir),
                 )
-                sleep(1)  # temporarily make things slower
                 model = train_model_func()
-                sleep(1)  # temporarily make things slower
                 self.tracker.mark_model_trained(
                     self.train_context.model_name,
                 )
-                sleep(1)  # temporarily make things slower
                 self.logger.info("Executed train_model_func successfully")
                 self._run_assertions(
                     model,
@@ -409,7 +405,6 @@ class ModelTrainer:
                 self.tracker.mark_model_saving(self.train_context.model_name)
                 self.logger.info(f"Saving model artifact {model} to model registry")
                 train.save_model(model, tracker=self.tracker)
-                sleep(1)  # temporarily make things slower
                 update_train_status(
                     self.client.model_catalog,
                     self.train_context.train_id,
--- a/layer/training/runtime/model_trainer.py
+++ b/layer/training/runtime/model_trainer.py
@@ -51,10 +51,17 @@ start_time_temp = None  # pylint: disable=C0103
 step_value = 0  # pylint: disable=C0103
 
 
-def get_gpu_metrics(logger: Logger) -> Dict[str, Dict[str, float]]:
+def _get_gpu_metrics(logger: Logger) -> Dict[str, Dict[str, float]]:
     metrics = {}
-    try:
-        for gpu in nvsmi.get_gpus():
+    gpu_present = nvsmi.is_nvidia_smi_on_path() is not None
+    if gpu_present:
+        try:
+            gpus = nvsmi.get_gpus()
+        except subprocess.CalledProcessError:
+            logger.info(
+                "Nvidia driver not running despite nvidia-smi being on the path. No GPU stats collected."
+            )
+        for gpu in gpus:
             metrics[gpu.id] = {
                 "utilisation": gpu.gpu_util,
                 "mem_utilisation": round(gpu.mem_util, 2),
@@ -63,14 +70,11 @@ def get_gpu_metrics(logger: Logger) -> Dict[str, Dict[str, float]]:
                 "id": gpu.id,
                 "name": gpu.name,
             }
-    except subprocess.CalledProcessError:
-        logger.info(
-            "Nvidia driver not running despite nvidia-smi being on the path. No GPU stats collected."
-        )
+
     return metrics
 
 
-def generate_metrics_dict(
+def _generate_system_metrics_dict(
     mem_used: float,
     mem_allocated: float,
     mem_used_percent: float,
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -348,6 +350,7 @@ skips = [
     "B101", # assert_used
     "B105", # hardcoded_password_string too noisy on defaults values empty string
     "B106", # hardcoded_password_funcarg too noisy on defaults values empty string
+    "B404", # import_subprocess
```
cihati, line 353: Instead of blank skip, please do where it's needed using `# nosec: import_subprocess`

layer/training/runtime/model_trainer.py
```diff
@@ -38,7 +46,71 @@ class TrainContextDataclassMixin:
...
+    try:
+        for gpu in nvsmi.get_gpus():
+            metrics[gpu.id] = {
+                "utilisation": gpu.gpu_util,
+                "mem_utilisation": round(gpu.mem_util, 2),
+                "mem_used": gpu.mem_used,
+                "mem_total": gpu.mem_total,
+                "id": gpu.id,
+                "name": gpu.name,
+            }
```
cihati, line 65: nit: I'd put `nvsmi.get_gpus()` into a variable and only wrap that line with try..except

layer/training/runtime/model_trainer.py
```diff
@@ -38,7 +46,71 @@ class TrainContextDataclassMixin:
     train_index: Optional[str] = None
 
 
+cpu_used_temp = None  # pylint: disable=C0103
+start_time_temp = None  # pylint: disable=C0103
+step_value = 0  # pylint: disable=C0103
+
+
+def get_gpu_metrics(logger: Logger) -> Dict[str, Dict[str, float]]:
```
cihati, line 54: nit: prepend all your private methods with `_`

layer/training/runtime/model_trainer.py
```diff
@@ -38,7 +46,71 @@ class TrainContextDataclassMixin:
...
+                "name": gpu.name,
+            }
+    except subprocess.CalledProcessError:
+        logger.info(
+            "Nvidia driver not running despite nvidia-smi being on the path. No GPU stats collected."
+        )
+    return metrics
+
+
+def generate_metrics_dict(
```
cihati, line 73: nit: `generate_system_metrics_dict`

layer/training/runtime/model_trainer.py
```diff
@@ -50,6 +122,77 @@ def __enter__(self) -> None:
     def get_working_directory(self) -> Path:
         pass
 
+    @staticmethod
+    def get_metrics(
```
cihati, line 126: nit: `generate_system_metrics`

layer/training/runtime/model_trainer.py
```diff
@@ -71,6 +215,35 @@ def get_working_directory(self) -> Path:
...
+        cpu_used = cpu_count * cpu_percent / 100
+
+        mem = psutil.virtual_memory()
+        mem_allocated = mem.total
+        mem_used = mem_allocated - mem.available
+        mem_utilisation = mem.percent
+
+        gpu_metrics = {}
+        gpu_present = nvsmi.is_nvidia_smi_on_path() is not None
+        if gpu_present:
```
cihati, line 233: nit: move this logic into `get_gpu_metrics`

layer/training/runtime/model_trainer.py
```diff
@@ -38,7 +46,71 @@ class TrainContextDataclassMixin:
     train_index: Optional[str] = None
 
 
+cpu_used_temp = None  # pylint: disable=C0103
```
cihati, line 49: why do you define this as a global var? same for `start_time_temp`

cihati (commented):
will take a deeper look, just a skim for now

layer/training/runtime/model_trainer.py
```diff
@@ -38,7 +46,75 @@ class TrainContextDataclassMixin:
...
+        "Memory Used (MB)": mem_used,
+        "Memory Allocated (MB)": mem_allocated,
+        "Memory Utilisation %": mem_used_percent,
+        "CPUs Used": cpus_used,
+        "CPUs Allocated": cpus_available,
+        "CPU Utilisation %": cpu_utilisation_percent,
+    }
+
+    for gpu in gpu_metrics:
+        metrics.update(
```
cihati, line 96: nit: put `gpu_metrics[gpu]["id"]` and `gpu_metrics[gpu]["name"]` just above here to make the statements below slightly cleaner. Also, could use f-strings instead of `format`.

layer/training/runtime/model_trainer.py
```diff
@@ -50,6 +126,72 @@ def __enter__(self) -> None:
...
+
+        def _get_mem_used() -> int:
+            return _read_value_from_file("/sys/fs/cgroup/memory/memory.usage_in_bytes")
+
+        def _get_mem_available() -> int:
+            return _read_value_from_file("/sys/fs/cgroup/memory/memory.limit_in_bytes")
+
+        def _get_used_percent(used: float, available: float) -> float:
+            if not used or not available:
+                print("System metric 0")
```
cihati, line 148: Is this intentional?

layer/training/runtime/model_trainer.py
```diff
@@ -167,6 +334,46 @@ def _train(
...
+                        logger=self.logger,
+                    )
+
+                    def _metrics_step_function(step: int) -> int:
+                        global step_value  # pylint: disable=invalid-name
+                        step += 1
+                        step_value += step
+                        return min(step, 15)
+
+                    if self.train_context.is_remote:
```
cihati, line 352: Instead of introducing `is_remote`, you could do smt like `not isinstance(self.train_context, LocalTrainContext)`. Also, this `if` block likely needs a comment.

layer/training/runtime/model_trainer.py
```diff
@@ -1,12 +1,19 @@
 import os
```
aleksmitov, line 1: Can we get the system metrics stuff outside of the `model_trainer` module, otherwise things get cluttered and more tightly coupled?

layer/training/runtime/model_trainer.py
aleksmitov: I think a better way would be to create a separate class holding these values, then attach an instance of this class somewhere, such as the train context. Then in your thread you pass this object as an argument, so that you can avoid working with globals

aleksmitov: About the step value, the logic is a bit not straightforward to grasp as an idea. Are you somehow trying to get the around the problem of us displaying metrics with step on the x axis, rather than time and emulating that through increasing the step value?
And if that's the case, this is kinda problematic because the step value on this graph would not be in sync with the ones on the other graphs. And worse, the step is generally intended to mean a specific intermediate state of the training of an ML model while here it would mean something else.

micheldlebeau: >I think a better way would be to create a separate class holding these values, then attach an instance of this class somewhere, such as the train context. Then in your thread you pass this object as an argument, so that you can avoid working with globals

I can look into doing that, it might take me a little while however.

micheldlebeau: >About the step value, the logic is a bit not straightforward to grasp as an idea. Are you somehow trying to get the around the problem of us displaying metrics with step on the x axis, rather than time and emulating that through increasing the step value?
And if that's the case, this is kinda problematic because the step value on this graph would not be in sync with the ones on the other graphs. And worse, the step is generally intended to mean a specific intermediate state of the training of an ML model while here it would mean something else.

It's true that right now the step annotations might make this confusing, however we're also planning to change how the graphs are displayed, which should address this.
Using steps is the only way to display a graph for now, so that's what we end up using.

layer/training/runtime/model_trainer.py
micheldlebeau: Removed the print and updated the logic slightly

layer/training/runtime/model_trainer.py
micheldlebeau: Much better, thanks!

layer/training/runtime/model_trainer.py
micheldlebeau: You might be able to save me a lot of time, can you advise where you think this should be moved to?

aleksmitov: For example you could create a `layer/training/runtime/metrics.py` and define functions and classes there and import them here. You could create this metrics holding class that we talked about in another thread and that could be the coupling point between the modules, apart from starting the actual metrics tracking in a thread

layer/training/runtime/system_metrics.py
```diff
@@ -0,0 +1,227 @@
...
+import psutil  # type: ignore
+
+from layer.clients.layer import LayerClient
+from layer.logged_data.log_data_runner import LogDataRunner
+
+
+class SystemMetrics:
+    cpu_used_temp = None
+    start_time_temp = None
+    step_value = 0
```
aleksmitov, line 19: You can define these inside the constructor, since different SystemMetrics instances would share the same values, which is not desirable: https://dzone.com/articles/python-class-attributes-vs-instance-attributes

layer/training/runtime/model_trainer.py
```diff
@@ -167,6 +169,20 @@ def _train(
...
+                    target=system_metrics.monitor_system_metrics,
+                    args=(
+                        lambda x: stop_system_metrics_thread,
+                        self.client,
+                        self.train_context.train_id,
+                        self.logger,
+                        isinstance(self.train_context, LocalTrainContext),
+                    ),
+                )
+                system_metrics_thread.start()
```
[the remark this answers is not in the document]
micheldlebeau: Contextmanager is definitely much nicer, thank you so much for all the feedback!

aleksmitov (approved):
👏

PR #153, opened 2022-06-23
Support for logging tensor and numpy images

mecevit:
Support for logging the following as images:
- torch.Tensor
- numpy.array in `HWC`, `CHW` and `HW` formats

Example:

```python
img = np.zeros((100, 100, 3))
img[:, :, 0] = np.arange(0, 10000).reshape(100, 100) / 10000
img[:, :, 1] = 1 - np.arange(0, 10000).reshape(100, 100) / 10000

layer.log({"np_image": layer.Image(img, format="HWC")})
```


fcakyon:
@mecevit tensorboard image constructor supports both float32 [0-1] and uint8 [0-255] images, is it the same in `layer.Image`?

from torch.utils.tensorboard source: https://github.com/pytorch/pytorch/blob/577e90ae9bf257040acb68da3626d9a64d07bf7a/torch/utils/tensorboard/summary.py#L431

mecevit:
> @mecevit tensorboard image constructor supports both float32 [0-1] and uint8 [0-255] images, is it the same in `layer.Image`?
> 
> from torch.utils.tensorboard source: https://github.com/pytorch/pytorch/blob/577e90ae9bf257040acb68da3626d9a64d07bf7a/torch/utils/tensorboard/summary.py#L431

Good catch, addressed here: https://github.com/layerai/sdk/pull/153/commits/06efc2c0649b5a70988d11a66caf267ecad14ae6

image

layer/logged_data/image.py
```diff
@@ -0,0 +1,121 @@
+from pathlib import Path
```
cihati, line 1: Please move this code under layer/contracts/logged_data.py

layer/logged_data/image.py
```diff
@@ -0,0 +1,121 @@
...
+        :param img: Supported image types are:
+        - torch.Tensor
+        - numpy.ndarray
+        - Image from path.Path
+        - PIL.Image
+        :param format: Sets the image format if the provided image is a numpy.ndarray.
+        Support formats are: `CHW`, `HWC`, `HW`
+        """
+        self.img = img
+        self.format = format
```
cihati, line 46: `raise ValueError` if `format` is unsupported -- instead of doing it in `_get_image_from_array` below.

layer/main.py
```diff
@@ -741,11 +743,12 @@ def log(
 
     **Images**
 
-    You can log images to track inputs, outputs, detections, activations and more. We support JPEG and PNG formats.
+    You can log images to track inputs, outputs, detections, activations and more. We support GIF, JPEG, PNG formats.
 
     Accepted Types:
     ``PIL.Image.Image``,
     ``path.Path``
+    ``layer.Image``
```
cihati, line 751: As we are not exposing `layer.Image`'s docs, this new PR's additions will not be very discoverable to the users. Either talk more about it here as part of `layer.log`, or make sure to send a PR for the docs separately in the monorepo.

mecevit: I will create a PR in monorepo

layer/logged_data/image.py
mecevit: Done

layer/__init__.py
```diff
@@ -7,6 +7,7 @@
 from .flavors.custom import CustomModel  # noqa
 from .global_context import current_project_name  # noqa
 from .logged_data.callbacks import KerasCallback, XGBoostCallback  # noqa
+from .logged_data.log_data_runner import Image  # noqa
```
cihati, line 10: Make sure to update this

cihati (approved):
Approving to unblock, but must address the last comment before merging.

layer/__init__.py
mecevit: Good catch. I wonder why my unit tests didn't cover that. Fixed.

PR #154, opened 2022-06-23
Add the executable tar creation logic for models

eminn:
This PR introduces an executable tar packaging under a feature flag called `LAYER_TAR_PACKAGING`.

In order to hide the feature we are copying the existing training code with slight differences until feature swap has been made. Then it would be the source of for both local and remote executions.


donatasm (approved):

layer/executables/tar.py
```diff
@@ -0,0 +1,104 @@
...
+    path: Path,
+    function: Callable[..., Any],
+    entrypoint: Path,
+    resources: Optional[List[Path]] = None,
+    pip_dependencies: Optional[List[str]] = None,
+) -> None:
+
+    # add pip dependencies needed by the extractor
+    pip_dependencies = pip_dependencies or []
+    pip_dependencies.extend(EXTRACTOR_PIP_DEPENDENCIES)
```
donatasm, line 25: To make sure we have reproducible tars between the different layer versions and local and remote builds, I suggest we include the layer package itself which was used to build the tar.

eminn: It somehow messed up on the CI when I change this and I don't want to waste time anymore on debugging that one, merging this and will address this later on

PR #159, opened 2022-06-24
Add dataset executions and make packaging explicit

eminn:


layer/decorators/dataset_decorator.py
```diff
@@ -142,27 +142,47 @@ def __init__(self, wrapped: Any, wrapper: Any, enabled: Any) -> None:
...
-                    account_name=current_project_full_name_.account_name,
+            dataset_definition = self.get_definition()
+            dataset_definition.package()
+            if is_feature_active("TAR_PACKAGING"):
+                import subprocess  # nosec: import_subprocess
+                import sys
+
+                subprocess.run(  # nosec: start_process_with_partial_path, subprocess_without_shell_equals_true
+                    [
+                        "sh",
```
aleksmitov, line 153: What about Windows users?

eminn: With the current form we are not supporting native Windows. We will look into how this plays with WSL.

volkangurel (approved):

PR #174, opened 2022-06-27
Only use numpy typing for type checking

volkangurel:
Only use numpy typing for type checking. This should resolve errors during training.

### DIFF
--- a/layer/contracts/logged_data.py
+++ b/layer/contracts/logged_data.py
@@ -4,12 +4,12 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Any, Dict, Union
 
 import numpy as np
-import numpy.typing as npt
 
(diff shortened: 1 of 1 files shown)


eminn (approved):

layer/contracts/logged_data.py
```diff
@@ -32,7 +32,9 @@ class Image:
 
     def __init__(
         self,
-        img: Union["PIL.Image.Image", Path, npt.NDArray[np.complex64], "torch.Tensor"],
+        img: Union[
+            "PIL.Image.Image", Path, "npt.NDArray[np.complex64]", "torch.Tensor"
+        ],
```
dpp23, line 37: There is another one on line 106 I think

volkangurel: Yep, fixed!

dpp23 (approved):

PR #179, opened 2022-06-28
Add a colab package validation test

dpp23:


build_scripts/colab-tests/pyproject.toml
```diff
@@ -0,0 +1,15 @@
+[tool.poetry]
```
dpp23, line 1: I've made this its own poetry package as it requires `poetry` as a dependency which:

a) is clashing with some of our other deps
b) is probably going to cause headaches for people using conda

build_scripts/colab-tests/validate_colab_dependencies.py
```diff
@@ -0,0 +1,85 @@
...
+                    f"Version check failed for '{pkg}'. Colab has {version_str}, but '{pkg_constraint['coming_from']}' requires {pkg_constraint['dep'].constraint}"
+                )
+    return violations
+
+
+if __name__ == "__main__":
+    parent_dir = pathlib.Path(__file__).parent
+    colab_packages = parse_colab_packages( parent_dir / "colab-packages.txt")
+    package_constraints = get_package_constraints(parent_dir.parent.resolve().parent.parent)
+    violations = find_violations(colab_packages, package_constraints) - IGNORED_VIOLATIONS
```
adrien-layer, line 79: insane syntax, but convenient

build_scripts/colab-tests/validate_colab_dependencies.py
```diff
@@ -0,0 +1,85 @@
...
+        )
+    )
+
+    for dependency in transitive_dependencies:
+        for transitive_dependency in dependency.package.requires:
+            constraints[transitive_dependency.name].append(
+                {"dep": transitive_dependency, "coming_from": dependency.name}
+            )
+
+    return constraints
```
adrien-layer, line 59: a bit worried about maintainability of this script if we use plain dicts, with "dep" etc... keys

can we introduce simple dataclasses instead?
maybe use some typing in method signatures just to give more guidance to future contributors

build_scripts/colab-tests/validate_colab_dependencies.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def find_violations(colab_packages, package_constraints):
+    violations = set()
+    for pkg, version_str in colab_packages.items():
+        version = Version.parse(version_str)
+        pkg_constraints = package_constraints[pkg]
+        for pkg_constraint in pkg_constraints:
+            if not pkg_constraint["dep"].constraint.allows(version):
```
adrien-layer, line 68: then this can be simplified to `pkg_constraint.allows(version)`

build_scripts/colab-tests/validate_colab_dependencies.py
```diff
@@ -0,0 +1,85 @@
...
+IGNORED_VIOLATIONS = {
+    "Version check failed for 'click'. Colab has 7.1.2, but 'flask' requires >=8.0",
+    "Version check failed for 'cloudpickle'. Colab has 1.3.0, but 'layer' requires >=2.0.0",
+    "Version check failed for 'grpcio'. Colab has 1.46.3, but 'grpcio-tools' requires >=1.47.0",
+    "Version check failed for 'humanize'. Colab has 0.5.1, but 'layer' requires >=3.11.0",
+    "Version check failed for 'itsdangerous'. Colab has 1.1.0, but 'flask' requires >=2.0",
+    "Version check failed for 'protobuf'. Colab has 3.17.3, but 'layer-api' requires >=3.19.4",
+    "Version check failed for 'pyarrow'. Colab has 6.0.1, but 'layer' requires 7.0.0",
+    "Version check failed for 'urllib3'. Colab has 1.24.3, but 'botocore' requires >=1.25.4,<1.27",
+}
```
adrien-layer, line 19: what is the plan for those?

build_scripts/colab-tests/colab-packages.txt
```diff
@@ -0,0 +1,400 @@
+# Generated by running `pip list` inside a frash colab runtime
```
adrien-layer, line 1: typo maybe? `fresh`

adrien-layer, line 1: do we want an entry on CONTRIBUTE for this process?

dpp23: I would leave it out for now.

build_scripts/colab-tests/validate_colab_dependencies.py
dpp23: No plan at the moment. I don't think we can eliminate them all. 

I see this test as linter - if we make a conscientious decision to ignore an error that is fine, but at least we will find out about these.

These are not (currently) fatal - we update them in Colab and things still work.

adrien-layer (approved):

PR #181, opened 2022-06-28
Upload executable packages

cihati:
LAY-3378

We don't have unit tests for ProjectRunner itself yet, so pushing automated tests for a separate PR when we have more pieces in place in the backend.


layer/projects/project_runner.py
```diff
@@ -5,11 +5,12 @@
 from typing import Any, Callable, List, Optional, Sequence
 
 import polling  # type: ignore
+import requests  # type: ignore
```
donatasm, line 8: `requests` is a transitive dependency, it should be added to `pyproject.toml`, if used directly. Can we do the upload with `aiohttp`, which sdk already [depends on](https://github.com/layerai/sdk/blob/a32b2903a070561f7e17775a6d2c737d3695d4c7/pyproject.toml#L55)?

cihati: @donatasm good point -- we also use it in layer/logged_data/log_data_runner.py . I should perhaps change it too (or add requests as a dep)

cihati: I made the change, but not 100% sure if I did it correctly tbh (Python async stuff is not my forte)

layer/projects/project_runner.py
```diff
@@ -160,6 +165,33 @@ def run(self, debug: bool = False, printer: Callable[[str], Any] = print) -> Run
...
+        session: aiohttp.ClientSession,
+    ) -> None:
+        with open(function.tar_path, "rb") as package_file:
+            presigned_url = client.executor_service_client.get_upload_path(
+                self.project_full_name, f"{function.asset_name}.tar"
+            )
+            await session.put(
+                presigned_url,
+                data=package_file,
+                timeout=aiohttp.ClientTimeout(total=None),
```
eminn, line 192: Is it intentional we don't enforce any timeout there ?

cihati: Not really, I c/p from another place (and we don't have timeout in either of those places)

layer/projects/project_runner.py
```diff
@@ -160,6 +165,33 @@ def run(self, debug: bool = False, printer: Callable[[str], Any] = print) -> Run
                     )
         return run
 
+    async def _upload_tar_packages(self, client: LayerClient) -> None:
+        trace_config = aiohttp.TraceConfig()
```
donatasm, line 169: `aiohttp.TraceConfig` is used in resource upload to report the progress, it is not required in this case.

pyproject.toml
```diff
@@ -68,6 +68,7 @@ protobuf  = ">=3.12.0"
 pyarrow = "7.0.0"
 pyjwt  = "^2.0.0"
 rich = ">=11"
+requests = ">=2.28.0"
```
donatasm, line 71: If `requests` are not used in favour of `aiohttp`,  this line should be removed.

dpp23: Can we reduce the lower bound? It is effectively latest or nothing at the moment and this might clash with anything else that has a lower upper bound. Colab has `2.23.0` for instance.

donatasm: In the future, we should stick to only one http client library.

donatasm (approved):

PR #189, opened 2022-06-30
Disable UI tracking based on an env variable

aleksmitov:
Fixes LAY-3425.
Will be used to turn off UI artifacts in execution logs for remote runs.


layer/tracker/progress_tracker.py
```diff
@@ -21,6 +22,11 @@
 from .output import get_progress_ui
 
 
+LAYER_DISABLE_TRACKING_UI = (
+    os.environ.get("LAYER_DISABLE_TRACKING_UI", "False").lower() == "true"
```
donatasm, line 26:
```
LAYER_DISABLE_TRACKING_UI = "LAYER_DISABLE_TRACKING_UI" in os.environ
```

donatasm:
Can we refactor so the code is not included with all the related dependencies instead?

PR #190, opened 2022-06-30
Use zipapp format to package executables

donatasm:
Implement function executable format provided by the built-in [zipapp](https://docs.python.org/3/library/zipapp.html) package instead of `tar`. The new format does not require the shell and other tools (the only requirement is `python` interpreter), thus can be executed in the environments like [distroless containers](https://github.com/GoogleContainerTools/distroless#why-should-i-use-distroless-images).

Usage example. Pack the function:

```python
from pathlib import Path
from layer.executables.packager import package_function

def simple():
    print("simple called")

package_function(simple, output_dir=Path("."))
```

The result of the snippet above is file called `simple`. The `simple` could be then executed directly in a shell:

```bash
./simple
simple called
```

or using `python` interpreter:

```bash
python ./simple
simple called
```

In both cases, for each invocation, a new temporary and isolated virtual environment is created in which the function is executed.


layer/executables/packager.py
```diff
@@ -0,0 +1,154 @@
...
+from typing import Any, Callable, List, Optional
+from uuid import uuid4
+
+import cloudpickle  # type: ignore
+
+
+def package_function(
+    function: Callable[..., Any],
+    resources: Optional[List[Path]] = None,
+    pip_dependencies: Optional[List[str]] = None,
```
cihati, line 18: Perhaps it's a concern of the users of this packager.py file, but https://docs.app.layer.ai/docs/sdk-library/pip-requirements-decorator also supports a file too (not just list of requirement strings)

layer/executables/packager.py
```diff
@@ -0,0 +1,154 @@
...
+            _package_resources(source, resources)
+
+        function_path = source / "function"
+        with open(function_path, mode="wb") as function_:
+            # register to pickle by value to ensure unpickling works anywhere
+            cloudpickle.register_pickle_by_value(sys.modules[function.__module__])
+            cloudpickle.dump(function, function_, protocol=pickle.DEFAULT_PROTOCOL)
+
+        # executable name is the function name (and not all callables always have a name)
+        executable_name = getattr(function, "__name__", f"callable-{uuid4()}")
```
cihati, line 46: can you please add a test for `(and not all callables always have a name)` aspect?

layer/executables/packager.py
```diff
@@ -0,0 +1,154 @@
...
+import tempfile
+import zipapp
+from pathlib import Path
+from typing import Any, Callable, List, Optional
+from uuid import uuid4
+
+import cloudpickle  # type: ignore
+
+
+def package_function(
```
cihati, line 15: can you please add a test where you call this function multiple times and make sure it works correctly? (there's quite a bit of resource cleanup supposed to happen, just wanna make sure all works well)

layer/executables/packager.py
```diff
@@ -0,0 +1,154 @@
...
+
+    for resource in resources:
+        if resource.is_file():
+            _copy_resource_file(resource)
+        elif resource.is_dir():
+            _walk_dir(resource)
+
+
+# depenedencies required by the runtime itself
+_runtime_requirements = ["cloudpickle==2.1.0"]
```
cihati, line 86: nit: move this above, make it all caps, fix the typo in the comment

cihati: What if `pip_requirements` also have `cloudpickle`?

layer/executables/packager.py
```diff
@@ -0,0 +1,154 @@
...
+                subprocess.check_call(pip_install)  # nosec
+
+                # extract resources
+                for entry in exec.infolist():
+                    if entry.filename.startswith("resources/"):
+                        with exec.open(entry, mode="r") as resource:
+                            target_path = (
+                                Path(".") / entry.filename[len("resources/") :]
+                            )
+                            target_path.parent.mkdir(parents=True, exist_ok=True)
```
cihati, line 129: should this be moved into the `if` block below?

layer/executables/packager.py
```diff
@@ -0,0 +1,154 @@
...
+            )
+
+        main_path = source / "__main__.py"
+        with open(main_path, mode="w", encoding="utf8") as main:
+            main.write(_loader_source())
+
+        if resources:
+            _package_resources(source, resources)
+
+        function_path = source / "function"
```
cihati, line 39: nit: I'd put `"function"` into a constant above.

layer/executables/packager.py
donatasm: Can you elaborate what exactly needs to be tested? All the files are stored under the temp dir, which gets deleted once the packaging is done.

layer/executables/packager.py
```diff
@@ -0,0 +1,153 @@
...
+        with open(function_path, mode="wb") as function_:
+            # register to pickle by value to ensure unpickling works anywhere
+            cloudpickle.register_pickle_by_value(sys.modules[function.__module__])
```
eminn, line 44: Are we sure about pickling by value, I know it wasn't introduced on this PR but based on the discussion on channel it wasn't a winner either.
It would be more in line with previous edition of the function pickling and we don't need to explicitly include cloudpickle as it should've been installed with layer. I think that is a requirement introduced by pickle by value somehow.

eminn:
Do you plan to remove the tar packaging and integrate new packager to the codebase on this PR ?

layer/executables/packager.py
donatasm: No, dirs are always created even though they're empty.

layer/executables/packager.py
donatasm: Not sure about  pickling by value either, but will remove, as less code is always simpler.

Regarding the `cloudpickle` dependency, I think it is better to keep it there explicitly to have a closed loop in serialising deserialising.

layer/executables/packager.py
donatasm: Not worth it, would require to redefine it in `_loader_source` anyway.

eminn (approved):

layer/executables/packager.py
donatasm: > Not sure about pickling by value either, but will remove, as less code is always simpler.

After digging into the docs, I think we should keep `register_pickle_by_value` call to have the executables less dependant on the external environments. In fact, removing `cloudpickle.register_pickle_by_value` fails all the tests in `test/unit/executables/test_packager` referencing `func_simple` and `func_with_dependencies`.

cihati: maybe expand the comment with more info for posterity?

layer/executables/packager.py
cihati: > All the files are stored under the temp dir, which gets deleted once the packaging is done.

basically wanna see if this holds if the package_function is called multiple times in a row.

cihati (approved):

layer/executables/packager.py
donatasm: > What if `pip_requirements` also have `cloudpickle`?

It will fail. I've added a test which is skipped for now and created a ticket to address, as it is dependency resolution issue, not packaging.

layer/executables/packager.py
donatasm: Got it, I added a test in https://github.com/layerai/sdk/pull/190/commits/2b51c65314afe60fe5ddce0911a210d1f4f97192.

PR #191, opened 2022-07-01
Downgrade h5py for M1 support

adrien-layer:


dpp23 (approved):

pyproject.toml
```diff
@@ -75,6 +75,7 @@ nvsmi = "^0.4.2"
 psutil = "^5.9.1"
 
 [tool.poetry.dev-dependencies]
+h5py = "<3.7.0"
```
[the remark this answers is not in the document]
adrien-layer: will add

.github/workflows/check.yml
```diff
@@ -45,6 +45,9 @@ jobs:
           make install
           poetry lock --no-update
           [ -z "$(git status --porcelain=v1 2>/dev/null)" ] || (echo "Lock file is not up to date, please run 'poetry lock --no-update'" && exit 1)
+      - name: Help on job failure
+        if: ${{ failure() }}
+        run: echo "Try running \"poetry cache clear --all pypi\" before \"poetry lock --no-update\""
```
dpp23, line 50: I don't think this will work - it will just be a collapsed step so no one will notice. Probably better to fold this into the `echo` in the previous step

adrien-layer: yeah... no one will notice this
Screenshot 2022-07-01 at 12 40 55

PR #197, opened 2022-07-07
Package function executable in layer run

donatasm:
* Package function executable in `layer.run`. Packaging is activated by feature flag `LAYER_EXECUTABLES=1`.
* Rename feature flag `LAYER_TAR_PACKAGING` to `LAYER_EXECUTABLES` (tars are going away in favour of zipapp).
* Remove tar packaging.


layer/projects/project_runner.py
```diff
@@ -166,23 +166,25 @@ def run(self, debug: bool = False, printer: Callable[[str], Any] = print) -> Run
...
 
-    async def _upload_tar_package(
+    async def _upload_executable_package(
         self,
         client: LayerClient,
         function: FunctionDefinition,
         session: aiohttp.ClientSession,
     ) -> None:
-        with open(function.tar_path, "rb") as package_file:
+        package_path = function.package()
```
eminn, line 183: We are already doing the packaging here https://github.com/layerai/sdk/blob/fb41d2d8e010ae29d1c5603df076b9fc0b6ca073/layer/projects/project_runner.py#L117 . I think one of them should be removed.

donatasm: Good catch, removed the packaging from the upload in https://github.com/layerai/sdk/pull/197/commits/b1bce8255bd5f81167ed0e689e94707c9359ceb9.

volkangurel (approved):
Looks good to me, just had a non-blocking comment/question.

layer/config/__init__.py
```diff
@@ -23,8 +23,7 @@
 from .config_manager import ConfigManager  # noqa
 
 
-def is_feature_active(feature_name: str) -> bool:
```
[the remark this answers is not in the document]
donatasm: I thought it is not worth keeping the generic code that is not used in a generic way yet.

PR #207, opened 2022-07-11
Fix asset statuses in sdk notebook UI 

aleksmitov:
Properly display asset statuses in the sdk's notebook UI.
Screen Shot 2022-07-11 at 19 32 11

### DIFF
--- a/layer/projects/progress_tracker_updater.py
+++ b/layer/projects/progress_tracker_updater.py
@@ -154,13 +154,10 @@ class ProgressTrackerUpdater:
             version_name = self.client.model_catalog.get_model_version(
                 ModelVersionId(value=str(task_id))
             ).name
-            self.tracker.mark_model_trained(
+            self.tracker.mark_model_saved(
                 name=model_name,
                 train_index=str(train_index),
                 version=version_name,
-            )
-            self.tracker.mark_model_saved(
-                name=model_name,
             )
         else:
             # TODO: alert for this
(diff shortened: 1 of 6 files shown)


layer/tracker/ui_progress_tracker.py
```diff
@@ -301,7 +304,7 @@ def mark_model_trained(
         self._update_asset(
             AssetType.MODEL,
             name,
-            status=AssetTrackerStatus.DONE,
+            status=AssetTrackerStatus.TRAINING,
```
volkangurel, line 307: Are you sure this is right? The function name is `mark_model_trained` but we're marking it as `training`.

layer/tracker/ui_progress_tracker.py
```diff
@@ -61,6 +62,9 @@ def track(self) -> Iterator["RunProgressTracker"]:
         with self._progress:
             self._init_tasks()
             yield self
+            time.sleep(
+                0.5
+            )  # Give some time so the latest metadata gets a chance to get rendered
```
volkangurel, line 67: Is this a limitation with rich UI? Is there a better option we can build down the road that doesn't require sleeps?

layer/projects/progress_tracker_updater.py
```diff
@@ -159,6 +159,9 @@ def _handle_task_succeeded(self, task: PBTask) -> None:
                 train_index=str(train_index),
                 version=version_name,
             )
+            self.tracker.mark_model_saved(
```
volkangurel, line 162: I think we should only send one event to the tracker here, either `trained` or `saved` but not both.

volkangurel (changes requested):

volkangurel (approved):

PR #209, opened 2022-07-12
Defer function metadata capture and packaging to runtime

donatasm:
Defer function metadata capture and packaging of dataset and model decorators to runtime. Now, all is done during the wrapper calls, which makes it difficult to have a single runtime for dataset builds and model trains. In case of executables, the decorators would only gather runtime info.

As a side effect, will close [LAY-3127](https://linear.app/layer/issue/LAY-3127/dataset-decorator-always-requiring-project).


maasg (approved):
This is very helpful to separate definition from execution :heavy_check_mark:

PR #216, opened 2022-07-15
Embed cloudpickle into executable

donatasm:
Embed cloudpickle `2.1.0` into the executable archive, so the package can load the user function by itself, without installing any external dependencies. Dependency installs are removed from the executable and will be handled by the runtime.


eminn (commented):
We should comply with the redistribution requirements https://github.com/cloudpipe/cloudpickle/blob/master/cloudpickle/cloudpickle.py#L14-L29

eminn (approved):

PR #219, opened 2022-07-18
Add group uuid to logged metrics

aleksmitov:
Adds the `LoggedMetricGroupId` to the logged metrics request. Including this id will allow to group metrics together, so that later the UI can visualize multiple metrics in the same chart.

### DIFF
--- a/layer/clients/logged_data_service.py
+++ b/layer/clients/logged_data_service.py
@@ -62,7 +62,7 @@ class LoggedDataClient:
         train_id: UUID,
         tag: str,
         points: List[ModelMetricPoint],
-        metric_group_uuid: UUID,
+        metric_group_id: UUID,
     ) -> ModelMetricId:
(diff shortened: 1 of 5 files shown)


layer/clients/logged_data_service.py
```diff
@@ -70,14 +75,19 @@ def get_logged_data(
         )
 
     def log_model_metric(
-        self, train_id: UUID, tag: str, points: List[ModelMetricPoint]
+        self,
+        train_id: UUID,
+        tag: str,
+        points: List[ModelMetricPoint],
+        metric_group_uuid: UUID,
```
cihati, line 82: nit: `metric_group_id`, to be consistent with `train_id` above and also `LoggedMetricGroupId`

cihati (approved):
Do we need to update the comment in `main.py`'s `log()`?

PR #222, opened 2022-07-18
Implement base function runtime

donatasm:
Skeleton contract between the function executable package and the runtime.

Basically, when started, the executable package asks for the runtime available by trying to call a global `__function_runtime` attribute. If one is provided, the executable will pass the loaded function to the runtime for execution.

All runtime implementations will extend the `BaseFunctionRuntime` class, which is a skeleton of a function execution workflow.

More concrete implementations will follow.

### DIFF
--- a/layer/executables/packager.py
+++ b/layer/executables/packager.py
@@ -6,16 +6,18 @@ import shutil  # nosec
 import sys
 import tempfile
 import zipapp
+import zipfile
+from dataclasses import dataclass
 from pathlib import Path
-from typing import Any, Callable, Generator, List, Optional
+from typing import Any, Callable, Generator, Optional, Sequence, Tuple
 
 from . import cloudpickle
 
 
 def package_function(
     function: Callable[..., Any],
-    resources: Optional[List[Path]] = None,
-    pip_dependencies: Optional[List[str]] = None,
+    resources: Optional[Sequence[Path]] = None,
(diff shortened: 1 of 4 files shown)


layer/executables/packager.py
```diff
@@ -102,41 +117,73 @@ def _walk_dir(dir: Path) -> None:
...
+            """Load the function runtime from the globals dictionary.
+            If none is provided, load the default runtime, which calls the function directly."""
+
+            # check for the runtime being provided from the outside
+            if "__function_runtime" in globals():
+                return __function_runtime  # type: ignore # noqa: F821 # pylint: disable=undefined-variable
+            else:
+
+                # return the default runtime, which calls the function directly
+                def _execute_function(function):  # type: ignore
```
maasg, line 144: `flatten()` :-)

test/unit/executables/test_runtime.py
```diff
@@ -0,0 +1,101 @@
...
+    def func_with_resources():
+        content = ""
+        for resource_path in resource_paths:
+            resource = Path(resource_path)
+            if resource.is_file():
+                with open(resource_path, "r") as f:
+                    content += f"file:{resource}={f.read()},"
+            elif resource.is_dir():
+                content += f"dir:{resource},"
+        print(f"content is '{content}'")
```
maasg, line 53: is this `print` still required?

maasg (approved):
:heavy_check_mark:

PR #227, opened 2022-07-19
Make Function class lists immutable

donatasm:
Make `Function` class lists immutable, as they should never change once created.


volkangurel (approved):

maasg (commented):
:+1:

PR #228, opened 2022-07-19
Use embedded cloudpickle

donatasm:
`cloudpickle` was embedded in https://github.com/layerai/sdk/pull/216 to make executables self extracting without installing any dependencies. This PR makes use of the embedded version everywhere in the sdk.


maasg (approved):
:heavy_check_mark:

PR #233, opened 2022-07-20
Package function asset name, type and other metadata

donatasm:
Package function asset name, type, sdk version used to package the function, as well as function serialiser name and version.


layer/executables/function.py
```diff
@@ -51,12 +56,38 @@ def pip_dependencies(self) -> Sequence[str]:
...
         return self._resources
 
+    @property
+    def metadata(self) -> Dict[str, Any]:
+        return {
+            "sdk": {
+                "version": layer.__version__,
+            },
+            "function": {
+                "serialiser": {
```
eminn, line 66: `serializer` ?

eminn (approved):

PR #248, opened 2022-07-27
Upload e2e-test logs

micheldlebeau:
Example at the bottom of this page: https://github.com/layerai/sdk/actions/runs/2744934691

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -76,8 +76,6 @@ jobs:
       matrix:
         python-version: [3.7, 3.8]
     runs-on: ubuntu-18.04
-    env:
-      SDK_E2E_TESTS_LOGS_DIR: build/e2e-home/stdout-logs/
     steps:
       - uses: actions/checkout@v2
       - run: git fetch origin ${{ github.event.pull_request.base.sha }}
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -76,6 +76,8 @@ jobs:
       matrix:
         python-version: [3.7, 3.8]
     runs-on: ubuntu-18.04
+    env:
+      SDK_E2E_TESTS_LOGS_DIR: build/e2e-home/stdout-logs/
     steps:
       - uses: actions/checkout@v2
       - run: git fetch origin ${{ github.event.pull_request.base.sha }}
@@ -125,5 +127,7 @@ jobs:
       - uses: actions/upload-artifact@v3
         with:
           name: e2e-test-logs
-          path: build/e2e-home/logs/
+          path: |
+            build/e2e-home/logs/*
+            build/e2e-home/stdout-logs/*
           retention-days: 30
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -122,3 +122,8 @@ jobs:
         run: |
           echo ${{ secrets.APP_API_KEY }} > .test-token
           make e2e-test
+      - uses: actions/upload-artifact@v3
+        with:
+          name: e2e-test-logs
+          path: build/e2e-home/logs/
+          retention-days: 30
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -126,7 +126,7 @@ jobs:
           make e2e-test
       - uses: actions/upload-artifact@v3
         with:
-          name: e2e-test-logs
+          name: e2e-test-logs (${{ matrix.python-version }})
           path: |
             build/e2e-home/logs/*
             build/e2e-home/stdout-logs/*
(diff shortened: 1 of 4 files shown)


.github/workflows/check.yml
```diff
@@ -122,3 +122,8 @@ jobs:
         run: |
           echo ${{ secrets.APP_API_KEY }} > .test-token
           make e2e-test
+      - uses: actions/upload-artifact@v3
+        with:
+          name: e2e-test-logs
```
dpp23, line 127: There are 2 parallel runs for different python versions. Is it a problem that we use the same artifact name for both?

dpp23 (commented):
We should also set `SDK_E2E_TESTS_LOGS_DIR` and capture the stdout logs as well which are otherwise hidden by `pytest-xdist`.

Also, is there a way to upload each file individually like CircleCI does? It will be a much better devex to be able to just open the file I care about rather than downloading an archive, unpacking it and only then finding the file I care about.

micheldlebeau:
> We should also set SDK_E2E_TESTS_LOGS_DIR and capture the stdout logs as well which are otherwise hidden by pytest-xdist.

Can do.

> Also, is there a way to upload each file individually like CircleCI does? It will be a much better devex to be able to just open the file I care about rather than downloading an archive, unpacking it and only then finding the file I care about.

I can look into that, but then what if we want to download all the files and search through them all? We end up having to manually click and download 32 files. 
Based on your experience, do you believe it's better to do it file by file as we never want to search through all the logs, or more generally we never need to look at more than a single log file?

dpp23:
> > We should also set SDK_E2E_TESTS_LOGS_DIR and capture the stdout logs as well which are otherwise hidden by pytest-xdist.
> 
> Can do.
> 
> > Also, is there a way to upload each file individually like CircleCI does? It will be a much better devex to be able to just open the file I care about rather than downloading an archive, unpacking it and only then finding the file I care about.
> 
> I can look into that, but then what if we want to download all the files and search through them all? We end up having to manually click and download 32 files. Based on your experience, do you believe it's better to do it file by file as we never want to search through all the logs, or more generally we never need to look at more than a single log file?

The way this used to work was:

1. See which test failed
2. Download the `stdout` log file for it - it has the test name, so easy to know which one you need
3. In `stdout` there was a log like `Session logs at: <a path to file>`
4. Based on the line from 3, one can download the exact session log file.


I see that 3 is now gone though - I think we should bring it back in a separate ticket. 

As for whether we want to upload an archive or multiple, if easy, I would say multiple files, if it isn't too slow. If not trivial though, an archive is good enough and we can adjust if we get feedback about it later.

.github/workflows/check.yml
```diff
@@ -76,6 +76,8 @@ jobs:
       matrix:
         python-version: [3.7, 3.8]
     runs-on: ubuntu-18.04
+    env:
+      SDK_E2E_TESTS_LOGS_DIR: build/e2e-home/stdout-logs/
```
dpp23, line 80: Can we put that in the makefile? We should capture these logs on local runs as well so people have easy access to them.

micheldlebeau:
> As for whether we want to upload an archive or multiple, if easy, I would say multiple files, if it isn't too slow. If not trivial though, an archive is good enough and we can adjust if we get feedback about it later.

The action out of the box essentially creates a file share with the name provided, and uploads all files matching the path to the file share. If we wanted a separate file share for each log file, we could probably use a matrix strategy, running the step once for each test log.
Now that would either require static log file names so that a static config could reference them, or we would need to dynamically generate the matrix, we could use logic like [this](https://brunoscheufler.com/blog/2021-10-09-generating-dynamic-github-actions-workflows-with-the-job-matrix-strategy), but it would require that ticket you mentioned for (3) which is now gone before this can work.

So for now I'd say we're probably better off with the archives.

CONTRIBUTING.md
```diff
@@ -250,6 +250,8 @@ Set these up with your IDE to have a smoother development experience and fewer f
 
 The final step after developing and testing your changes locally is to submit a pull request and get your contribution merged back into `layerai/sdk`. Please follow the instructions in the GitHub template when creating your PR and fix any status checks that are failing.
 
+To help debug E2E test issues, network logs are captured, zipped and available in the Summary page of each Check GitHub Action page after the E2E tests are executed.
```
dpp23, line 253: Might be worth adding a line where they go locally.

layer/main.py
```diff
@@ -739,6 +739,8 @@ def log(
 
     Logs arbitrary data associated with a model train or a dataset build into Layer backend.
 
+    Extra line to trigger e2e-tests.
+
```
dpp23, line 743:
```suggestion
```

dpp23 (approved):

PR #251, opened 2022-07-27
Add parameters to make e2e-test

dpp23:


micheldlebeau (approved):

CONTRIBUTING.md
```diff
@@ -229,6 +229,20 @@ make e2e-test
 You will be asked for your key which will be stored for subsequent runs in `.test-token`.
 You can find the test logs under `build/e2e-home/logs` and also the standard output generated during tests under `build/e2e-home/stdout-logs`.
 
+##### Run a subset of the e2e-tests
+You can pass `E2E_TEST_SELECTOR` to `make e2e-test` to select a subset of the tests to run. In this example we are running a single test:
```
micheldlebeau, line 233: How would you pass a subset with more than a single test? Comma separated?

dpp23: There is quite an extensive list of supported methods - https://docs.pytest.org/en/latest/how-to/usage.html#specifying-tests-selecting-tests

micheldlebeau: Makes sense, looking at this document alone it wasn't clear if the whole range of pytest selectors could be used, or if this was limited to a single test.
```suggestion
You can pass `E2E_TEST_SELECTOR` to `make e2e-test` to select a subset of the tests to run. Use the [standard pytest syntax](https://docs.pytest.org/en/latest/how-to/usage.html#specifying-tests-selecting-tests) to specify which tests to invoke.  
In this example we are running a single test:
```

PR #252, opened 2022-07-27
Dimitar/inf 1205 print link to datadog traces in sdk ci

dpp23:


.github/workflows/release.yml
```diff
@@ -93,9 +93,14 @@ jobs:
           datadog_site: datadoghq.eu
       - name: Run e2e tests
         if: steps.changed-files.outputs.any_changed == 'true'
-        run: |
-          echo ${{ secrets.APP_API_KEY }} > .test-token
-          make e2e-test
+        run: bash build_scripts/ci/e2e_tests.sh ${{ secrets.APP_API_KEY }}
+      - uses: actions/upload-artifact@v3
```
micheldlebeau, line 97: Missed that one, thank you

build_scripts/ci/e2e_tests.sh
```diff
@@ -0,0 +1,23 @@
...
+
+start_time=$(date +%s%3N)
+set +e
+make e2e-test
+exit_code=$?
+set -e
+end_time=$(date +%s%3N)
+
+echo ""
+echo -e "${RED}Logs${NC}: https://app.datadoghq.eu/logs?from_ts=$start_time&to_ts=$end_time&index=%2A&live=false&query=stack%3Aapp"
```
micheldlebeau, line 17: nit: Why do we use red here?
These get printed no matter what the exit code is, so maybe this should be a neutral colour instead, like blue?

dpp23: As we discussed in person, changed the colour to red on failure and green on success.

micheldlebeau (approved):

PR #263, opened 2022-08-03
INF-1211 Make virtualenv setup easier

almg80:
Makes M1 macs work with other conda versions other than miniforge3
Splits Makefile to reduce complexity of one big file
Add a create and remove environment command to help users have a working conda env

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,15 +1,6 @@
-ROOT_DIR := $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
-INSTALL_STAMP := .install.stamp
-E2E_TEST_HOME := $(ROOT_DIR)/build/e2e-home
-E2E_TEST_SELECTOR := test/e2e
-E2E_TEST_PARALLELISM := 16
-TEST_TOKEN_FILE := .test-token
-POETRY := $(shell command -v poetry 2> /dev/null)
-IN_VENV := $(shell echo $(CONDA_DEFAULT_ENV)$(CONDA_PREFIX)$(VIRTUAL_ENV))
-CONDA_ENV_NAME := $(shell echo $(CONDA_DEFAULT_ENV))
-UNAME_SYS := $(shell uname -s)
-UNAME_ARCH := $(shell uname -m)
-REQUIRED_POETRY_VERSION := 1.1.14
+include include.Makefile
+include environment.Makefile
+
 
 .DEFAULT_GOAL:=help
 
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -1,15 +1,6 @@
...
-E2E_TEST_SELECTOR := test/e2e
-E2E_TEST_PARALLELISM := 16
-TEST_TOKEN_FILE := .test-token
-POETRY := $(shell command -v poetry 2> /dev/null)
-IN_VENV := $(shell echo $(CONDA_DEFAULT_ENV)$(CONDA_PREFIX)$(VIRTUAL_ENV))
-CONDA_ENV_NAME := $(shell echo $(CONDA_DEFAULT_ENV))
-UNAME_SYS := $(shell uname -s)
-UNAME_ARCH := $(shell uname -m)
-REQUIRED_POETRY_VERSION := 1.1.14
+include Makefile.envs
```
almg80, line 1: Going through pr, think would be better named as `Makefile.environments`

Makefile
```diff
@@ -118,23 +109,10 @@ publish: ## Publish to PyPi - should only run in CI
 	$(POETRY) publish --build --username $(PYPI_USER) --password $(PYPI_PASSWORD)
 	$(POETRY) version $(CURRENT_VERSION)
 
-.PHONY: clean
```
almg80, line 121: Moved to `Makefile.envs`

Makefile.envs
```diff
@@ -0,0 +1,49 @@
...
+	$(CONDA_EXE) env update -p build/$(PROJECT_NAME) -f $<
+	@echo
+	@echo "New conda env created."
+else
+	$(error Unsupported Environment. Please use conda)
+endif
+endif
+endif
+
+.PHONY: delete_environment ## Delete the virtual (conda) environment
```
almg80, line 31: Not yet added a Linux, since still debating if we should

environment.M1.yml
```diff
@@ -0,0 +1,9 @@
+name: sdk
+channels:
+  - default
+  - apple
+  - conda-forge
+dependencies:
+  - python=3.8.13
```
almg80, line 7: Should we instead of hardcode versions, generate it dynamically from inside Makefile and `.python-version`?

Makefile
```diff
@@ -118,23 +110,10 @@ publish: ## Publish to PyPi - should only run in CI
...
-	@rm -rf dist/
-	@rm -f $(INSTALL_STAMP)
-	@find . -type f -name '*.pyc' -delete
-	@find . -type d -name "__pycache__" | xargs rm -rf {};
-	@echo 'done.'
-
 .PHONY: deepclean
 deepclean: clean ## Resets development environment including test credentials and venv
 	@rm -rf `poetry env info -p`
+	@rm -rf build/$(PROJECT_NAME)
```
dpp23, line 116: deepclean should probably just nuke all of `build`

dpp23 (approved):

PR #266, opened 2022-08-04
Test against colab before merging sdk

micheldlebeau:
For now, the image is built and pushed locally, manually.
The automation for automatically getting new dependencies from Colab and building the new image will be handled separately.
Tested on M1 mac and on Linux (via CI).


.github/workflows/check.yml
```diff
@@ -70,6 +70,14 @@ jobs:
         run: make check
       - name: Run check-colab-violations
         run: make check-colab-violations
+  colab-test:
```
dpp23, line 73: Need to add to `release.yml` as well.

include.mk
```diff
@@ -0,0 +1,22 @@
...
+UNAME_ARCH := $(shell uname -m)
+REQUIRED_POETRY_VERSION := 1.1.14
+PROJECT_NAME := sdk
+COLAB_TEST_HOME := $(ROOT_DIR)/build/colab-test
+COLAB_IMAGE_BUILD_STAMP := .image-built.stamp
+DOCKER_IMAGE_NAME = layerco/colab-lite
+ifdef CI
+	DOCKER_RUN = @docker run -v $(shell pwd):/usr/src/app:ro -v $(shell pwd)/dist:/usr/src/app/dist:rw --rm --platform=linux/amd64 -e LAYER_API_KEY=$(shell cat .test-token) --name colab-test $(DOCKER_IMAGE_NAME)
+else
+	DOCKER_RUN = @docker run -v $(shell pwd):/usr/src/app:ro -v $(shell pwd)/dist:/usr/src/app/dist:rw --rm --platform=linux/amd64 -e LAYER_API_KEY=$(shell cat .test-token) --name colab-test -it $(DOCKER_IMAGE_NAME)
```
dpp23, line 20: Why is the command different if not in CI?

test/colab/colab-test.mk
```diff
@@ -0,0 +1,31 @@
+.PHONY: colab-test-internal
```
dpp23, line 1: Can we move the `DOCKER_RUN` definition to this file? It is easier to read without jumping files and it is only used here.

include.mk
```diff
@@ -0,0 +1,22 @@
...
+POETRY := $(shell command -v poetry 2> /dev/null)
+IN_VENV := $(shell echo $(CONDA_DEFAULT_ENV)$(CONDA_PREFIX)$(VIRTUAL_ENV))
+CONDA_ENV_NAME := $(shell echo $(CONDA_DEFAULT_ENV))
+UNAME_SYS := $(shell uname -s)
+UNAME_ARCH := $(shell uname -m)
+REQUIRED_POETRY_VERSION := 1.1.14
+PROJECT_NAME := sdk
+COLAB_TEST_HOME := $(ROOT_DIR)/build/colab-test
+COLAB_IMAGE_BUILD_STAMP := .image-built.stamp
+DOCKER_IMAGE_NAME = layerco/colab-lite
```
dpp23, line 16: Can we put a concrete tag?

Makefile
```diff
@@ -1,6 +1,6 @@
-include include.Makefile
-include environment.Makefile
-
+include include.mk
```
micheldlebeau, line 1: Use standard extension for makefiles

Makefile
```diff
@@ -61,6 +61,20 @@ endif
 	LAYER_DEFAULT_PATH=$(E2E_TEST_HOME) SDK_E2E_TESTS_LOGS_DIR=$(E2E_TEST_HOME)/stdout-logs/ $(POETRY) run python build_scripts/sdk_login.py $(TEST_TOKEN_FILE)
 	LAYER_DEFAULT_PATH=$(E2E_TEST_HOME) SDK_E2E_TESTS_LOGS_DIR=$(E2E_TEST_HOME)/stdout-logs/ $(POETRY) run pytest $(E2E_TEST_SELECTOR) -s -n $(E2E_TEST_PARALLELISM) -vv $(DATADOG_ARGS)
 
+.PHONY: colab-test
+colab-test: ## Run colab test against image pulled from dockerhub
+# Catching sigint/sigterm to forcefully interrupt run on ctrl+c
+	@/bin/bash -c "trap \"trap - SIGINT SIGTERM ERR; echo colab-test cancelled by user; exit 1\" SIGINT SIGTERM ERR; $(MAKE) colab-test-internal"
```
micheldlebeau, line 67: When running locally this allows interrupting a run easily rather than having to manually run `docker rm -f colab-test` in a separate session.

Makefile
```diff
@@ -61,6 +61,20 @@ endif
...
+
+.PHONY: colab-test-local
+colab-test-local: ## Run colab test against image built locally
+# Catching sigint/sigterm to forcefully interrupt run on ctrl+c
+	@/bin/bash -c "trap \"trap - SIGINT SIGTERM ERR; echo colab-test cancelled by user; exit 1\" SIGINT SIGTERM ERR; $(MAKE) colab-test-internal-local"
+
+.PHONY: colab-test-push
+colab-test-push: colab-test-build ## Push image built locally to dockerhub
+	@docker push $(DOCKER_IMAGE_NAME)
```
micheldlebeau, line 76: Only executed locally, manually at the moment.

test/colab/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+    && apt-get autoremove -y
+
+ENV CPLUS_INCLUDE_PATH=/usr/include/gdal
+ENV C_INCLUDE_PATH=/usr/include/gdal
+
+COPY build/colab-test/requirements-fixed.txt /tmp/requirements-fixed.txt
+RUN --mount=type=cache,mode=0755,target=/root/.cache/pip pip install --no-dependencies -r /tmp/requirements-fixed.txt
+WORKDIR /usr/src/app
+
+CMD ["./test/colab/test.sh"]
```
micheldlebeau, line 23: This image ends up being 8GB uncompressed and 3GB compressed, FYI

test/colab/colab-test.mk
```diff
@@ -0,0 +1,31 @@
...
+colab-test-internal: $(TEST_TOKEN_FILE) $(COLAB_TEST_HOME)/test_import_login_init.ipynb colab-test-pull
+	$(DOCKER_RUN)
+
+.PHONY: colab-test-internal-local
+colab-test-internal-local: $(TEST_TOKEN_FILE) $(COLAB_TEST_HOME)/test_import_login_init.ipynb colab-test-build
+	$(DOCKER_RUN)
+
+.PHONY: colab-test-pull
```
micheldlebeau, line 9: Always run this, if there has been no changes it takes no time at all.  
Simpler than having some extra logic to keep track of whether the image has already been pulled, and we don't have to manage the scenario where there's a new version.

test/colab/colab-test.mk
```diff
@@ -0,0 +1,31 @@
...
+colab-test-pull:
+	@docker pull $(DOCKER_IMAGE_NAME)
+
+.PHONY: colab-test-build
+colab-test-build: $(COLAB_IMAGE_BUILD_STAMP)
+$(COLAB_IMAGE_BUILD_STAMP): $(COLAB_TEST_HOME)/requirements-fixed.txt test/colab/Dockerfile
```
micheldlebeau, line 15: Only build once as it's time consuming, unless the Dockerfile or requirements have changed, in which case we rebuild it.

test/colab/test.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -eo pipefail
+
+trap 'wait;' SIGINT SIGTERM
```
micheldlebeau, line 4: Trap mechanism to ensure that interruptions stop the docker run as soon as it's received.

test/colab/test.sh
```diff
@@ -0,0 +1,10 @@
...
+
+. $HOME/.poetry/env
+poetry build --format wheel
```
dpp23, line 7: Can we just mount the wheel in the container and do the build ourside? Feels odd to do it inside the container.

test/colab/Dockerfile
dpp23, line 23: This file is not actually in the image right? Can we not set CMD and instead pass this to the container when we run it.

test/colab/colab-test.mk
```diff
@@ -0,0 +1,31 @@
+.PHONY: colab-test-internal
+colab-test-internal: $(TEST_TOKEN_FILE) $(COLAB_TEST_HOME)/test_import_login_init.ipynb colab-test-pull
+	$(DOCKER_RUN)
+
+.PHONY: colab-test-internal-local
+colab-test-internal-local: $(TEST_TOKEN_FILE) $(COLAB_TEST_HOME)/test_import_login_init.ipynb colab-test-build
+	$(DOCKER_RUN)
+
+.PHONY: colab-test-pull
+colab-test-pull:
```
dpp23, line 10: Doesn't docker try to pull automatically if an image is not there?

test/colab/colab-test.mk
```diff
@@ -0,0 +1,31 @@
...
+	@mkdir -p $(COLAB_TEST_HOME)
+# Fix requirements.txt to make them compatible with running on M1 mac/linux, outside of colab itself.
+	@./test/colab/fix-requirements.sh
+
+$(COLAB_TEST_HOME)/test_import_login_init.ipynb: test/colab/test_import_login_init.ipynb
+	@mkdir -p $(COLAB_TEST_HOME)
+# VSC saves notebooks with `null` in place of `0` for `execution_count`, e.g. `"execution_count": null,`.\
+`nbconvert` crashes on these, so we fix that here. 
+	@cp test/colab/test_import_login_init.ipynb $(COLAB_TEST_HOME)/test_import_login_init.ipynb
+	@sed -i'' -e "s/\"execution_count\": null,/\"execution_count\": 0,/g" $(COLAB_TEST_HOME)/test_import_login_init.ipynb
```
dpp23, line 31: I think we want that in `make format`, not when we run the tests as this notebook is committed in the code.

test/colab/fix-requirements.sh
```diff
@@ -0,0 +1,19 @@
...
+# Ensure setuptools version lower than 58 is used, as 58 and above break the install due to use_2to3 not being available
+sed -i'' -e "1s/^/setuptools<58\n/" build/colab-test/requirements-fixed.txt
+
+# Remove Tensorflow libraries due to AVX missing on M1 mac
+sed -i'' -e "s/kapre==/# kapre==/" build/colab-test/requirements-fixed.txt
+sed -i'' -e "s/keras==/# keras==/" build/colab-test/requirements-fixed.txt
+sed -i'' -e "s/tensorflow==/# tensorflow==/" build/colab-test/requirements-fixed.txt
+
+# Remove non-existing versions
+sed -i'' -e "s/google-colab @/# google-colab @/" build/colab-test/requirements-fixed.txt
```
dpp23, line 12: Instead of removing, should we try installing from [here](https://github.com/googlecolab/colabtools) ?

test/colab/test_import_login_init.ipynb
```diff
@@ -0,0 +1,70 @@
...
+        "layer.login_with_api_key(os.environ.get(\"LAYER_API_KEY\"))"
+      ]
+    },
+    {
+      "cell_type": "code",
+      "execution_count": 0,
+      "metadata": {},
+      "outputs": [],
+      "source": [
+        "layer.init(\"test_colab\")"
```
dpp23, line 44: This has the side effect of creating a new project so it has different behaviour the first time it is run vs any other.

If we really want to create projects, then we should create a randomly named one and then delete after.


Doing just login is good enough for now though.

test/colab/test_import_login_init.ipynb
```diff
@@ -0,0 +1,70 @@
...
+        },
+        "id": "vufPU6M0On4-",
+        "outputId": "1e572b8b-ce7f-4907-8c42-54b638a07789"
+      },
+      "outputs": [],
+      "source": [
+        "import layer\n",
+        "import os\n",
+        "\n",
+        "layer.login_with_api_key(os.environ.get(\"LAYER_API_KEY\"))"
```
dpp23, line 35: Shall we call `layer.login()` after?

include.mk
micheldlebeau: We need -it to be able to interrupt locally, however on CI -t leads to the `the input device is not a TTY` error

include.mk
micheldlebeau: Currently we implicitly use `latest` as there's no point in testing against anything else, so do you want us to specify `:latest`, or do you want to actually version the image when pushing and figure out which image to use when pulling?

test/colab/Dockerfile
micheldlebeau: Not sure I get what the difference would be?

test/colab/colab-test.mk
micheldlebeau: It does, but if the image is already there it won't pull again. This ensures that the latest image is always used. It also takes no time at all if the latest version is already present on the machine.

test/colab/colab-test.mk
micheldlebeau: No, we create a copy of the notebook under the `build` folder, and format that notebook, so this doesn't get committed.

test/colab/fix-requirements.sh
micheldlebeau: I can look into that, are you thinking of always pulling from master here, or manually pinning to the latest commit when getting the requirements from colab?

test/colab/test_import_login_init.ipynb
micheldlebeau: I believe what happens on the backend doesn't actually matter here, as the SDK in colab will go through the same logic nonetheless?

environment.mk
```diff
@@ -64,6 +64,8 @@ clean: ## Resets development environment.
 	@find . -depth -type d -name '*.egg-info' -not -path "*build/$(PROJECT_NAME)/*" -delete
 	@rm -rf dist/
 	@rm -f $(INSTALL_STAMP)
+	@rm -f $(COLAB_IMAGE_BUILD_STAMP)
+	@rm -rf $(ROOT_DIR)/build/
```
almg80, line 68: Should we nuke `build` here? We already nuke on a `deepclean` if we remove here all conda env and ipython will be removed on MacOS forcing new env to be created.

micheldlebeau: Sure

test/colab/colab-test.mk
```diff
@@ -0,0 +1,31 @@
...
+
+.PHONY: colab-test-internal-local
+colab-test-internal-local: $(TEST_TOKEN_FILE) $(COLAB_TEST_HOME)/test_import_login_init.ipynb colab-test-build
+	$(DOCKER_RUN)
+
+.PHONY: colab-test-pull
+colab-test-pull:
+	@docker pull $(DOCKER_IMAGE_NAME)
+
+.PHONY: colab-test-build
```
almg80, line 20: Should not be `PHONY` right? We want to track `$(COLAB_IMAGE_BUILD_STAMP)`

micheldlebeau: We do track it, in the line right underneath:
```
.PHONY: colab-test-build
colab-test-build: $(COLAB_IMAGE_BUILD_STAMP)
```
It's just used as an alias to make the `colab-test-internal-local` command easier to understand, but we could just as well remove `.PHONY` and use `$(COLAB_IMAGE_BUILD_STAMP)` instead, if we wanted to be less verbose. WDYT?

test/colab/colab-test.mk
```diff
@@ -0,0 +1,38 @@
...
+colab-test-build: $(COLAB_IMAGE_BUILD_STAMP)
+$(COLAB_IMAGE_BUILD_STAMP): $(COLAB_TEST_HOME)/requirements-fixed.txt test/colab/Dockerfile
+	@docker build -t $(DOCKER_IMAGE_NAME) -f test/colab/Dockerfile .
```
almg80, line 23: On a random ubuntu I was testing which did not had buildkit enabled it errors out complaining about it. We could `DOCKER_BUILDKIT=1` or maybe just add as a requirements to use it locally.

micheldlebeau: Fair, I assumed that was now the default, but maybe it's not, thanks!

test/colab/colab-test.mk
almg80: I'm ok with keeping it as it is, thanks for explaining it!

include.mk
dpp23: Maybe worth a comment in that case?

test/colab/test.sh
dpp23: I don't think the python version used to build the wheel matters. We build it with 3.8 when we push to pypi and that works on all python versions we support.

test/colab/Dockerfile
dpp23: The difference is that we don't bake anything in the image that is not related to colab.

For instance now if I decide to rename this file, I need to rebuild the whole image.

test/colab/colab-test.mk
dpp23: In that case, you can just pass `--pull always` to `docker run`

test/colab/colab-test.mk
dpp23: But when I checked out your branch and edited the notebook I got a diff with that exact same problem.

test/colab/fix-requirements.sh
dpp23: I guess `master` should be good as this is probably what they use when they build their images.

test/colab/test_import_login_init.ipynb
dpp23: > I believe what happens on the backend doesn't actually matter here, as the SDK in colab will go through the same logic nonetheless?

Sorry, I don't follow what you mean. `layer.init` has a side effect as it will create a project. We should not have side effects from tests.

test/colab/Dockerfile
micheldlebeau: CMD provides a default which can be overwritten.  
If you want to rename the file you can update the docker run command and add `./test/colab/new-test-file.sh` to the end of it to point to the new file, as you say.  

The point of a default is to make sense in most cases, and for now we're only planning to use this image for this one purpose.

test/colab/colab-test.mk
micheldlebeau: We have 2 usecases:
- Sometimes we want to build the image locally, run the image we just built locally, and also potentially push it from local to dockerhub
- Sometimes we just want to pull the latest from dockerhub and run that, e.g. in CI, or most devs on local as it's faster to pull and build from scratch.

If we update the run command with your suggestion, then for the first scenario we would build the image locally, and then pull the remote image and run that instead of the image built locally.

test/colab/colab-test.mk
dpp23: Can no longer reproduce this.

test/colab/test_import_login_init.ipynb
micheldlebeau: Gotcha, my point what that whether the project was created or not, the behaviour on the SDK side would be the same when it comes to dependencies loaded and code path.  
I've removed that part of the test to follow your "no side effect allowed" rule.

test/colab/colab-test.mk
dpp23: Ok, thanks for explaining. This is a bit convoluted, but we can revisit when we do the automatic builds.

test/colab/Dockerfile
dpp23: I think it is better to keep the image agnostic from the test code instead of being tightly coupled. I personally think of this image as a CI executor, just like the ones we use in CircleCI - it has the right environment for us to add the test code and execute. I see them as two separated components we will be mainteining and integrating separately.

We will have to break this tight coupling when we start building the image nightly and likely move to a separate repository.

It is also quite confusing that the CMD references a file that isn't actually in the image.

I am ok with keeping it as it is in the context of this PR, but I feel like we will have to break that coupling soon after.

micheldlebeau: I don't believe how we build the image will affect how we use it. When we automate the creation of the image, it will still only be used for this test, initially.  
If down the road we have another usecase for the image and end up using it as a general purpose executor, then sure, we might have to decouple the image from the tests, and maybe even create a different image building on top of this one, like we do for the executors, but let's not optimise prematurely.

For now this is a single purpose image, let's just focus on making sure it does what it's supposed to do well first. 😃 

Let's talk about the automated build process separately, as I'm sure we'll disagree on that too!

.github/workflows/release.yml
```diff
@@ -50,6 +50,22 @@ jobs:
       - name: Check package loads without dev dependencies
         run: make check-package-loads
 
+  colab-test:
```
dpp23, line 53: Shall we block release if this fails?

dpp23 (approved):

.github/workflows/release.yml
micheldlebeau: Yes, absolutely, thanks!

PR #267, opened 2022-08-04
INF-1204 - Add a make jupyter command to SDK

almg80:
- Jupyter installed as a dev dependency.
- Create an ipython profile and add a file to always run autoreload extension
- Fix auto `make help` generation
- Add `make jupyter` that will start jupyter notebook in the terminal with editable layer package installed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,10 @@ include environment.Makefile
 
 .DEFAULT_GOAL:=help
 
+define get_python_package_version
+  $(1)==$(shell $(POETRY) show $1 --no-ansi --no-dev | grep version | awk '{print $$3}')
+endef
+
 define autoreloadpy
 from IPython import get_ipython
 ipython = get_ipython()
@@ -12,10 +16,6 @@ ipython.magic("load_ext autoreload")
 ipython.magic("autoreload 2")
 endef
 export autoreloadpy
-
-define get_python_package_version
-  $(1)==$(shell $(POETRY) show $1 --no-ansi --no-dev | grep version | awk '{print $$3}')
-endef
 
 install: check-poetry prereq-$(UNAME_SYS) $(INSTALL_STAMP) ## Install dependencies
 $(INSTALL_STAMP): pyproject.toml poetry.lock .python-version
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -4,6 +4,15 @@ include environment.Makefile
...
 .DEFAULT_GOAL:=help
 
+define autoreloadpy
+from IPython import get_ipython
+ipython = get_ipython()
+
+ipython.magic("load_ext autoreload")
+ipython.magic("autoreload 2")
+endef
+export autoreloadpy
```
dpp23, line 14: Can we move out of the main makefile?

Makefile
```diff
@@ -71,7 +82,7 @@ lint: $(INSTALL_STAMP) ## Run all linters
 	$(POETRY) run isort --check-only .
 	$(POETRY) run black --check . --diff
 	$(POETRY) run flake8 .
-	$(POETRY) run pylint  --recursive yes .
+	$(POETRY) run pylint  --recursive yes . --ignore=build
```
dpp23, line 85: Can we move this to `[tool.pylint.MASTER]` in pyproject.toml? there is already an exclusion list there.

Makefile
```diff
@@ -15,6 +24,8 @@ ifdef IN_VENV
 else
 	$(POETRY) install --remove-untracked
 endif
+	@poetry run ipython profile create --ipython-dir=build/ipython
```
dpp23, line 27: Is creating the profile idempotent? As in if it exists already this won't fail the second time we run it?

environment.Makefile
```diff
@@ -54,6 +54,11 @@ ifeq ($(UNAME_ARCH), arm64)
 	rm -fr build/$(PROJECT_NAME)
 endif
 
+.PHONY: jupyter
+jupyter: install ## Start a jupyter notebook with editable layer package
+	@pip install -e .
```
dpp23, line 59: When you do `poetry run` that should already have the package installed in editable mode. Are you sure we need `pip install`? Will this even be the correct pip if poetry is using a virtual env?

Makefile
almg80: We can invoke it multiple times, it won't fail

environment.Makefile
almg80: During tests was using MacOS and `conda`, was installing via pip to conda venv and then running jupyter directly instead of via poetry. After testing and making it work in Linux that `pip` is a leftover and not needed. Thanks!

dpp23 (approved):

PR #271, opened 2022-08-05
Remove colab dependency violation check

dpp23:
Now that we have colab like test, this old approximation is no longer necessary as the new tests should catch any fatal conflicts.


micheldlebeau:
Could we keep it for a week or two to see if the new solution is actually much better at catching issues?

dpp23:
@micheldlebeau shall we merge this one now? We've had one false positives in this test already on a PR

micheldlebeau (approved):

PR #272, opened 2022-08-05
Introduce conda decorator

eminn:


layer/decorators/__init__.py
```diff
@@ -1,3 +1,4 @@
+from .conda_decorator import conda  # noqa
```
volkangurel, line 1: Can we import this in `layer/__init__.py` as well so that users can use it as `@layer.conda`?

layer/executables/packager.py
```diff
@@ -37,6 +40,10 @@ def package_function(
         # include cloudpickle itself in the executable
         _copy_cloudpickle_package(source)
 
+        if conda_env is not None:
+            conda_environment_file_path = source / "environment.yml"
```
donatasm, line 44: Is this conventional name for conda? If not, maybe rename to something more specific - `conda_env.yml`?

donatasm (approved):

layer/executables/packager.py
eminn: Yes, it's called like that https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#creating-an-environment-from-an-environment-yml-file

PR #274, opened 2022-08-05
Add organization account e2e test

adrien-layer:
We add first organization e2e test for:
- building a dataset
- training a model

(we reuse existing test methods to improve maintainability and parity checks with personal accounts)

- We create only one organization account but one project for each test.
- The test creates and cleans up the organization account.

### DIFF
--- a/test/e2e/conftest.py
+++ b/test/e2e/conftest.py
@@ -1,9 +1,10 @@
+import asyncio
 import logging
 import os
 import uuid
 from contextlib import contextmanager
 from pathlib import Path
-from typing import Any, Callable, Iterator, Optional, Tuple
+from typing import Any, Callable, Iterator, Tuple
 
 import ddtrace
 import pytest
@@ -11,7 +12,6 @@ import pytest
 import layer
 from layer.clients.layer import LayerClient
 from layer.config import ClientConfig, Config, ConfigManager
-from layer.contracts.accounts import Account
 from layer.contracts.fabrics import Fabric
 from layer.contracts.projects import Project
 from layer.projects.project_runner import ProjectRunner
@@ -81,7 +81,7 @@ def _extract_module_and_test_name(fixture_request: Any) -> Tuple[str, str]:
 
 
 def _slugify_name(src: str) -> str:
-    return _cut_off_prefixing_dirs(src).replace("[", "-").replace("]", "")
+    return _cut_off_prefixing_dirs(src).replace("[", "-").replace("]", "").lower()
 
 
 def _cut_off_prefixing_dirs(module_name: str) -> str:
@@ -90,7 +90,7 @@ def _cut_off_prefixing_dirs(module_name: str) -> str:
     return module_name.split(".")[-1]
 
 
-@pytest.fixture()
+@pytest.fixture(autouse=True)
 def _write_stdout_stderr_to_file(capsys, request) -> Iterator[Any]:
     """
     Since we parallelize pytest tests via xdist, stdout and stderr are swallowed.
@@ -136,48 +136,27 @@ def _write_stdout_stderr_to_file(capsys, request) -> Iterator[Any]:
                 f.write(captured.err)
 
 
-@pytest.fixture()
-async def config(_write_stdout_stderr_to_file: Any) -> Config:
+# https://github.com/tortoise/tortoise-orm/issues/638#issuecomment-830124562
+# so that async fixtures with session scope can be run
+@pytest.fixture(scope="class")
+def event_loop():
+    return asyncio.get_event_loop()
+
+
+@pytest.fixture(scope="class")
+async def config() -> Config:
     return await ConfigManager().refresh()
 
 
-@pytest.fixture()
+@pytest.fixture(scope="class")
 async def client_config(config: Config) -> ClientConfig:
     return config.client
 
 
-@pytest.fixture()
+@pytest.fixture(scope="class")
 def client(client_config: ClientConfig) -> Iterator[LayerClient]:
     with LayerClient(client_config, logger).init() as client:
         yield client
-
-
-@pytest.fixture()
-def initialized_organization_account(
-    client: LayerClient, request: Any
-) -> Iterator[Account]:
-    org_account_name = pseudo_random_account_name(request)
-    account = client.account.create_organization_account(org_account_name)
-
-    yield account
-    _cleanup_account(client, account)
-
-
-def _cleanup_account(client: LayerClient, account: Account):
-    client.account.delete_account(account_id=account.id)
-
-
-@pytest.fixture()
-def initialized_organization_project(
-    client: LayerClient, initialized_organization_account: Account
-) -> Iterator[Project]:
-    account_name = initialized_organization_account.name
-    # no need for pseudo randomness here, already done at the organization account level
-    project_name = "organization-project"
-    project = layer.init(f"{account_name}/{project_name}", fabric=Fabric.F_XSMALL.value)
-
-    yield project
-    _cleanup_project(client, project)
 
 
 @pytest.fixture()
--- a/test/e2e/conftest.py
+++ b/test/e2e/conftest.py
@@ -42,6 +42,8 @@ def pytest_sessionstart(session):
 
     org_account = _read_organization_account_from_test_session_config()
     if org_account:
+        # This is unexpected, so we cleanup and raise an exception in order to address the underlying issue
+        _cleanup_organization_account()
         raise Exception(
             f"pytest_sessionstart test session config already setup with account {org_account.name} {org_account.id}"
         )
@@ -64,6 +66,10 @@ def _track_session_counts_and_cleanup() -> Iterator:
     if open_sessions < 1:
         _cleanup_organization_account()
         _cleanup_test_session_config()
+        if open_sessions < 0:
+            raise Exception(
+                f"unexpected negative number of open sessions {open_sessions}"
+            )
 
 
 def _cleanup_test_session_config() -> None:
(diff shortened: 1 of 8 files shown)


adrien-layer:
failing because we need a new token with new org account claim

need to relogin / refresh token after org account creation

marta-tatiana (approved):

test/e2e/test_organization_function_run.py
```diff
@@ -0,0 +1,82 @@
...
+    @pytest.fixture(scope="class")
+    def initialized_organization_account(
+        self, client: LayerClient, request: Any
+    ) -> Iterator[Account]:
+        org_account_name = pseudo_random_account_name(request)
+        account = client.account.create_organization_account(org_account_name)
+
+        # We need a new token with permissions for the new org account
+        # TODO LAY-3583 LAY-3652
+        refresh_login(force=True)
```
dpp23, line 42: Is this going to cause races between different tests trying to refresh the login?

For guest logins, we had to do a per-test config.json. see https://github.com/layerai/sdk/blob/main/test/e2e/conftest.py#L148

test/e2e/test_organization_function_run.py
```diff
@@ -0,0 +1,82 @@
...
+# We use a wrapping Test class so we create only one organization
+# until we handle newly acquired permissions (after org creation)
+# better
+class TestOrganization:
+    @pytest.fixture(scope="class")
+    def initialized_organization_account(
+        self, client: LayerClient, request: Any
+    ) -> Iterator[Account]:
+        org_account_name = pseudo_random_account_name(request)
+        account = client.account.create_organization_account(org_account_name)
```
dpp23, line 38: Are all users allowed to create an organization account?

test/e2e/test_organization_function_run.py
adrien-layer: yes i wanted to test before going down that route, but it is likely we'll need to

test/e2e/test_organization_function_run.py
adrien-layer: atm yes, it will evolve later when we add restrictions

dpp23: Not a blocker but one thing to keep in mind so we update the docs when we get there - the current assumption is that anyone can run e2e-tests with their personal account. If the user needs special permissions, then we will need to make it explicit that `e2e-tests` are only for internal use.

We can just add a flag/selector to skip them as well.

adrien-layer: do you think it is worth adding that flag now to skip them when choosing to?

im thinking when time comes it will be part of the task to do that

test/e2e/test_organization_function_run.py
adrien-layer: actually come to think of it: the permissions needed there are additive, the only risk i see is that we invalidate the token of other tests but other tests actually do a config refresh (without force) as part of their fixture chaining, so i think this should be okay

test/e2e/test_organization_function_run.py
dpp23: No, let's leave it for later when it is actually a problem. Just wanted to mention it for future consideration, no need to do this now.

test/e2e/test_organization_function_run.py
adrien-layer: this is irrelevant now that we use the pytest_sessionstart hook

test/e2e/conftest.py
```diff
@@ -30,24 +33,202 @@ def pytest_sessionstart(session):
...
+
+    Since we parallelize tests, and fixture teardowns happen after pytest_sessionfinish hook, we need
+    this autoUse fixture to clean up the global test session.
+    """
+    yield
+    session_finished = _is_test_session_finished()
+    if not session_finished:
+        return
+    _cleanup_organization_account(client)
+    _cleanup_test_session_config()
```
dpp23, line 146: Are we sure this will get executed at least once?

test/e2e/conftest.py
```diff
@@ -30,24 +33,202 @@ def pytest_sessionstart(session):
...
+
+
+def pytest_sessionfinish(session):
+    """
+    WARNING This hook runs BEFORE fixture teardown.
+
+    We use it to set a flag across processes that the global session has finished,
+    so we can safely cleanup shared setup. Otherwise this shared setup can be deleted
+    while some tests have not started yet due to parallelism.
+    """
```
dpp23, line 52: I don't quite follow what is the sequence here and what we are trying to solve by this file signaling.

Is `pytest_sessionfinish` called on the main process before all tests are finished?

test/e2e/conftest.py
adrien-layer: it should yes since the flag is set by `pytest_sessionfinish` with '`a`' append mode

what could happen though is that `pytest_sessionfinish` suddenly starts being executed after fixture teardowns, so i can add an attempt to delete the org account also inside the hook

test/e2e/conftest.py
adrien-layer: yes kinda, meaning fixture teardowns are executed after `pytest_sessionfinish` . i should improve the explanation here then

test/e2e/conftest.py
dpp23: Are fixtures cleaned up after every test or at the end of the run?

What will happen if the tests get split like this:

w0 - orgTest, regularTest
w1 - orgTest, regularTest

?

If the tear down happens after `orgTest`, that will be before `pytest_sessionfinish` so we will never clean up.

adrien-layer: i dont understand the question sorry. currently there is an autoUse fixture that will be run after every test, so one single test teardown will be the one to execute the session cleanup

dpp23: oh, I see, so that cleanup is attempted after every test regardless of whether it actually uses the org or not.

adrien-layer: yes indeed, this is why I call it `_test_session_teardown`, kinda like our own hook for our distributed use case.

i thought `pytest_sessionfinish` was that but it isnt behaving as i expected

test/e2e/conftest.py
adrien-layer: found this https://github.com/pytest-dev/pytest/issues/7484, last commit uses it

test/e2e/conftest.py
adrien-layer: using this https://github.com/pytest-dev/pytest/issues/7484 now

dpp23 (approved):

adrien-layer:
org accounts are actually not cleaned up

seems pytest_sessionfinish simply does not run with xdist

im gonna use a file lock now because these hooks are too unpredictable
also this is what xdist suggests https://pypi.org/project/pytest-xdist/#making-session-scoped-fixtures-execute-only-once

adrien-layer:
hooray, now organization account is being cleaned up properly. also tested locally.

after merge i will monitor this is reliable

test/e2e/conftest.py
```diff
@@ -1,53 +1,250 @@
...
+
+    _write_organization_account_to_test_session_config(org_account)
+
+
+@pytest.fixture(scope="session", autouse=True)
+def _track_session_counts_and_cleanup() -> Iterator:
+    """
+    This will run for every pytest session.
+    Since we optionally use xdist to parallelize test runs,
+    we need extra logic tp do track parallel pytest sessions and do teardown once there are no more.
```
adrien-layer, line 59: fix typo

dpp23:
```suggestion
    we need extra logic to do track parallel pytest sessions and do teardown once there are no more.
```

test/e2e/conftest.py
```diff
@@ -1,53 +1,250 @@
...
+
+
+def _cleanup_organization_account() -> None:
+    async def refresh() -> Config:
+        return await ConfigManager().refresh()
+
+    loop = asyncio.get_event_loop()
+    config = loop.run_until_complete(refresh())
+    client = LayerClient(config.client, logger)
+    account = _read_organization_account_from_test_session_config()
```
adrien-layer, line 166: do this first so we avoid setting up the client for nothing

test/e2e/conftest.py
```diff
@@ -1,53 +1,250 @@
...
         ddtrace.patch_all()
 
+    if xdist.is_xdist_worker(session):
+        return
+
+    org_account = _read_organization_account_from_test_session_config()
+    if org_account:
+        raise Exception(
+            f"pytest_sessionstart test session config already setup with account {org_account.name} {org_account.id}"
+        )
```
dpp23, line 47: If I get in this scenario, how do I unblock myself? Would it be better if we just try to delete it and move on? Maybe error if we fail?

test/e2e/conftest.py
```diff
@@ -1,53 +1,250 @@
...
+def _track_session_counts_and_cleanup() -> Iterator:
+    """
+    This will run for every pytest session.
+    Since we optionally use xdist to parallelize test runs,
+    we need extra logic tp do track parallel pytest sessions and do teardown once there are no more.
+    """
+    _increment_session_count()
+    yield
+    open_sessions = _decrement_session_count()
+    if open_sessions < 1:
```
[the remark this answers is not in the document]
adrien-layer: i thought about this and thought we'd still need to cleanup but we do want to be aware yes, will add sthg

test/e2e/conftest.py
adrien-layer: this was to prevent a bad setup and should fail early. but i should delete then before failing so we don't leak.
i would still have this fail early instead of silently ignore this illegal state

PR #275, opened 2022-08-05
Train models in layer_runtime

donatasm:
Train models in `layer_runtime` module. Invocation similar to datasets:

```bash
 python -m layer.executables.layer_runtime --project=layer/layer-runtime-55 ./train_model
```

### DIFF
--- a/layer/executables/layer_runtime.py
+++ b/layer/executables/layer_runtime.py
@@ -69,7 +69,7 @@ class LayerFunctionRuntime(BaseFunctionRuntime):
             self._create_model(function_output.name, func)
         else:
             raise LayerFunctionRuntimeError(
-                f"unsupported function output type: {function_output.type}"
+                f"missing function output type: {function_output.type}"
             )
 
     def _layer_init(self) -> None:
--- a/layer/executables/layer_runtime.py
+++ b/layer/executables/layer_runtime.py
@@ -1,5 +1,7 @@
+import contextlib
 import logging
 import os
+import tempfile
 import uuid
 from argparse import ArgumentParser
 from dataclasses import dataclass
@@ -10,13 +12,16 @@ from layerapi.api.ids_pb2 import ProjectId
 
 import layer
 from layer.clients.data_catalog import DataCatalogClient
+from layer.clients.model_catalog import ModelCatalogClient
 from layer.clients.project_service import ProjectServiceClient
 from layer.config.config import ClientConfig
 from layer.config.config_manager import ConfigManager
 from layer.contracts.asset import AssetPath, AssetType
 from layer.contracts.fabrics import Fabric
+from layer.contracts.models import TrainStorageConfiguration
 from layer.executables.packager import FunctionPackageInfo
 from layer.executables.runtime import BaseFunctionRuntime
+from layer.flavors.utils import get_flavor_for_model
 from layer.global_context import current_project_full_name, set_has_shown_update_message
 
 
@@ -60,6 +65,8 @@ class LayerFunctionRuntime(BaseFunctionRuntime):
         function_output = _get_function_output(self._package_info.metadata)  # type: ignore
         if function_output.is_dataset:
             self._create_dataset(function_output.name, func)
+        elif function_output.is_model:
+            self._create_model(function_output.name, func)
         else:
             raise LayerFunctionRuntimeError(
                 f"unsupported function output type: {function_output.type}"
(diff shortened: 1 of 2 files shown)


layer/executables/layer_runtime.py
```diff
@@ -97,6 +104,50 @@ def _create_dataset(self, name: str, func: Callable[..., Any]) -> None:
         )
         data_catalog.store_dataset(func(), uuid.UUID(build_response.id.value))
 
+    def _create_model(self, name: str, func: Callable[..., Any]) -> None:
+        model = func()
+        model_flavor = get_flavor_for_model(model)
+        if model_flavor is None:
+            raise LayerFunctionRuntimeError("unsupported model flavor")
```
maasg, line 111:
```suggestion
            raise LayerFunctionRuntimeError("missing model flavor")
```

layer/executables/layer_runtime.py
```diff
@@ -97,6 +104,50 @@ def _create_dataset(self, name: str, func: Callable[..., Any]) -> None:
...
+            model_catalog = ModelCatalogClient.create(self._client_config, self._logger)  # type: ignore
+            project = current_project_full_name()
+            asset_path = AssetPath(
+                name,
+                AssetType.MODEL,
+                org_name=project.account_name,  # type: ignore
+                project_name=project.project_name,  # type: ignore
+            )
+            source_digest = _get_function_source_digest(self._package_info.metadata)  # type: ignore
+            display_fabric = Fabric.F_LOCAL.value  # the fabric to display in the UI
```
maasg, line 123: why is this `LOCAL` ? I guess it has to be a parameter we pass to the function. This is only for reporting purposes. The fabric should be 'consumed' by the executor service to create the right deployment.

layer/executables/layer_runtime.py
```diff
@@ -97,6 +104,50 @@ def _create_dataset(self, name: str, func: Callable[..., Any]) -> None:
...
+            model_catalog.complete_model_train(
+                model_train_id, model_flavor.PROTO_FLAVOR
+            )
+            model_storage_config = model_catalog.get_model_train_storage_configuration(
+                model_train_id
+            )
+            _upload_model_artifacts(Path(model_dir), model_storage_config)
+            model_catalog.store_training_metadata(
+                asset_name=asset_path.asset_name,
+                description="",
```
maasg, line 143: where should we get this description from?

layer/executables/layer_runtime.py
```diff
@@ -143,6 +194,32 @@ def _get_function_output(metadata: Dict[str, Any]) -> _FunctionOutput:
     return _FunctionOutput(type, name)
 
 
+def _get_function_source_digest(metadata: Dict[str, Any]) -> str:
+    source_digest = metadata.get("function", {}).get("source", {}).get("digest")
```
maasg, line 198: poor's man `flatmap` ;-)

layer/executables/layer_runtime.py
```diff
@@ -143,6 +194,32 @@ def _get_function_output(metadata: Dict[str, Any]) -> _FunctionOutput:
...
+    if source_digest is None:
+        raise LayerFunctionRuntimeError("function metadata missing source digest")
+    return source_digest
+
+
+def _upload_model_artifacts(model_dir: Path, config: TrainStorageConfiguration) -> None:
+    import boto3
+
+    s3_config = {
+        "aws_access_key_id": config.credentials.access_key_id,
```
maasg, line 208: Where are we going to get these from? Is there a `.layer/config` in the runner? Maybe populated by the `login()` function?

maasg (commented):
There might be a couple of points to align with the executor interface.

layer/executables/layer_runtime.py
donatasm: Due to legacy reasons, it has to be passed. I had to put any value here. We would need to pass it from the runtime. Same applies to the datasets. Will address in a separate PR.

layer/executables/layer_runtime.py
donatasm: `poor's man ~ pythonic` 🤣

layer/executables/layer_runtime.py
donatasm: > Where are we going to get these from?

The credentials are generated in the backend. Access by the API key.

layer/executables/layer_runtime.py
```diff
@@ -97,6 +104,50 @@ def _create_dataset(self, name: str, func: Callable[..., Any]) -> None:
         )
         data_catalog.store_dataset(func(), uuid.UUID(build_response.id.value))
 
+    def _create_model(self, name: str, func: Callable[..., Any]) -> None:
```
eminn, line 107: The Context object created and set as active context to make logging and other stuff work because that's the way they figure out which function execution are happening. How are we going to solve that problem ?

layer/executables/layer_runtime.py
donatasm: https://linear.app/layer/issue/LAY-3646/read-fabric-value-from-the-env-in-layer-runtime

layer/executables/layer_runtime.py
donatasm: From the `dataset` or `model` decorators. Will be mapped in a separate PR. https://linear.app/layer/issue/LAY-3647/capture-and-store-function-descriptions-in-the-runtime

layer/executables/layer_runtime.py
donatasm: Haven't looked at it yet, but I feel it could be solved by not using the global state of the process.

maasg (approved):
:heavy_check_mark:

maasg:
bors merge

PR #288, opened 2022-08-09
Allow passing in args and kwargs for model training

volkangurel:
Allow passing in args and kwargs for model training


mecevit (approved):

maasg:
Adding @donatasm as this might impact the work on the new runtime.

eminn:
Is this for ray PoC ? If that's the case the model trainer we have there in `layer.executables.model` and that needs to be updated.

eminn (approved):

PR #301, opened 2022-08-10
Do not redirect stdout to rich Console

aleksmitov:
This fixes most likely a race condition when stdout is redirected to the rich Console.

### DIFF
--- a/layer/tracker/output.py
+++ b/layer/tracker/output.py
@@ -64,4 +64,7 @@ def get_progress_ui() -> Progress:
         SpinnerColumn(finished_text=":white_heavy_check_mark:"),
         AssetColumn(),
         console=Console(),
+        # Google Colab runtime crashes occasionally if below params are set to true
+        redirect_stderr=False,
+        redirect_stdout=False,
     )


layer/tracker/output.py
```diff
@@ -64,4 +64,7 @@ def get_progress_ui() -> Progress:
         SpinnerColumn(finished_text=":white_heavy_check_mark:"),
         AssetColumn(),
         console=Console(),
+        # Google Colab runtime crashes occasionally if below params are set to true
+        redirect_stderr=False,
```
eminn, line 68: Could we add some e2e test for colab which verifies that this actually fixes the issue ? @micheldlebeau recently introduced a test setup for colab so he might help on that one.

micheldlebeau: The colab test currently doesn't go any further than `layer.login()`, but it could easily be extended to have e2e-tests.
Take a look inside `test/colab/` and let me know if you have any questions.

eminn: Thanks @micheldlebeau from there I think we could have a basic tensorflow model train with verbose logging and see if it crashes or not.

aleksmitov: Can't really do that because 1) I can only reproduce the crash with a tensorflow script that runs for ~10mins with a GPU and 2) We can't run tensorflow on the colab image at the moment.

eminn (approved):

PR #306, opened 2022-08-11
Remove unnecessary guard and allow override from controller

adrien-layer:


donatasm (approved):

test/e2e/conftest.py
```diff
@@ -40,13 +40,6 @@ def pytest_sessionstart(session):
     if xdist.is_xdist_worker(session):
         return
 
-    org_account = _read_organization_account_from_test_session_config()
-    if org_account:
-        # This is unexpected, so we cleanup and raise an exception in order to address the underlying issue
-        _cleanup_organization_account()
```
dpp23, line 46: We don't want to try to clean up still?

dpp23 (approved):

test/e2e/conftest.py
adrien-layer: good point, added back

PR #307, opened 2022-08-11
Add function assertions

eminn:


layer/executables/layer_runtime.py
```diff
@@ -105,13 +108,17 @@ def _create_dataset(self, name: str, func: Callable[..., Any]) -> Any:
             ProjectId(value=str(self._project_id)), name, display_fabric
         )
         dataset = func()
+        dataset = check_and_convert_to_df(dataset)
```
donatasm, line 111: This check should probably not be there. Is it required for assertions? If it is just to ensure the `DataFrame` is passed here, I suggest not to solve this problem in this PR. This is the scope of https://linear.app/layer/issue/LAY-3464/add-validation-to-datasetmodel-decorators-to-check-function-signatures to solve the issue globally.

layer/executables/layer_runtime.py
```diff
@@ -155,6 +162,20 @@ def _create_model(self, name: str, func: Callable[..., Any]) -> Any:
...
+        failed_assertions = []
+        for assertion in reversed(assertions):
+            try:
+                assertion.function(object_under_test)
+            except Exception:
+                failed_assertions.append(assertion)
+        if len(failed_assertions) > 0:
+            raise LayerFailedAssertionsException(failed_assertions)
+        else:
+            print(f"{len(assertions)} assertions run successfully.")
```
donatasm, line 177: Looks like debug leftover, remove.

donatasm (approved):

layer/executables/layer_runtime.py
eminn: Similar to the supported model flavors we need to make sure that we'd getting a dataframe from the function so that assertions could operator on that one.

Issue #244, opened 2022-07-23
logging experiments without a decorator and without returning model file


fcakyon:
## Summary

1. While logging anything to layer.ai, a dataset or model (consider only these two decorators for simplicity) decorator is needed to wrap a function. However, I need a way of logging parameters/metrics more dynamically. 

For instance, considering a `LayerTracker` class as below (modified version of [huggingface/accelerate wandb tracker](https://github.com/huggingface/accelerate/blob/6c4edc362f85e79fac143c41a189bc90cb42e29e/src/accelerate/tracking.py#L177)), user cannot change the model decorator argument name dynamically as it is hardcoded:

```python
from accelerate.tracking import GeneralTracker

class LayerTracker(GeneralTracker):
    """
    A `Tracker` class that supports `layer`. Should be initialized at the start of your script.
    Args:
        run_name (`str`):
            The name of the experiment run.
        kwargs:
            Additional key word arguments passed along to the `layer.init` method.
    """

    name = "layer"
    requires_logging_directory = False

    def __init__(self, run_name: str, **kwargs):
        self.run_name = run_name

        access_token = kwargs.get("access_token", None)
        if access_token is not None:
            layer.login_with_access_token(access_token)
        self.project: layer.contracts.projects.Project = layer.init(project_name=self.run_name, **kwargs)
        logger.info(f"Initialized Layer project {self.run_name}")
        logger.info("Make sure to log any initial configurations with `self.store_init_configuration` before training!")

    def store_init_configuration(self, values: dict):
        """
        Logs `values` as hyperparameters for the run. Should be run at the beginning of your experiment.
        Args:
            values (Dictionary `str` to `bool`, `str`, `float` or `int`):
                Values to be stored as initial hyperparameters as key-value pairs. The values need to have type `bool`,
                `str`, `float`, `int`, or `None`.
        """
        layer.log({"config": values})
        logger.info("Stored initial configuration hyperparameters to Layer")

    @model("model")
    def log(self, values: dict, step: Optional[int] = None, **kwargs):
        """
        Logs `values` to the current run.
        Args:
            values (Dictionary `str` to `str`, `float`, `int` or `dict` of `str` to `float`/`int`):
                Values to be logged as key-value pairs. The values need to have type `str`, `float`, `int` or `dict` of
                `str` to `float`/`int`.
            step (`int`, *optional*):
                The run step. If included, the log will be affiliated with this step.
            kwargs:
                Additional key word arguments passed along to the `layer.log` method.
        """
[...]

PR #308, opened 2022-08-12
Make models optional to return from the model decorator

volkangurel:
Make models optional to return from the model decorator.

This should address https://github.com/layerai/sdk/issues/244 by letting users return `None` from their training runs without impacting Layer successfully finishing a training run and collecting metadata.


eminn (commented):
Did we check how the UI would be broken by the missing fields from a successful train ?

volkangurel:
Yep, the UI looks good:
image

volkangurel:
Sorry, here's the one without a returned model. Still looks good :
image

eminn (approved):

fcakyon:
@volkangurel is there a way to log model via logging api? Something like `layer.log_model(model)` or `layer.log(model)`?

fcakyon:
got my answer its `save_model` :+1:

PR #312, opened 2022-08-15
Create Layer Logger for PL

mecevit:
LayerLogger implementation for Pytorch Lightning


layer/logged_data/loggers.py
```diff
@@ -0,0 +1,357 @@
+from argparse import Namespace
+from pathlib import Path
+from typing import Any, Dict, Generator, List, Mapping, MutableMapping, Optional, Union
+
+import torch
+
+import layer
+
+
+class LayerLogger("pytorch_lightning.loggers.logger.Logger"):
```
volkangurel, line 10: Can we call this class `PytorchLightningLayerLogger`?

layer/logged_data/loggers.py
```diff
@@ -0,0 +1,357 @@
+from argparse import Namespace
```
volkangurel, line 1: Can we make this a file named `pytorch_lightning.py` under the `loggers/` directory?

layer/logged_data/loggers.py
mecevit: Done

layer/logged_data/loggers.py
mecevit: @cihati won't like it but done 😈

volkangurel (approved):

PR #313, opened 2022-08-15
Organize main functions

volkangurel:
Organize main functions by splitting up the 1000 line `main.py` into a directory.

Also add an `sdk_function` decorator to do the version checking correctly for all functions.


layer/logged_data/log_data_runner.py
```diff
@@ -73,7 +73,7 @@ def log(
             elif isinstance(value, bool):
                 self._log_boolean(tag=tag, bool_val=value)
             elif isinstance(value, (int, float)):
-                if self._train_id and epoch is not None:
```
cihati, line 76: Metrics are only meant for models, so this was correct as is. Please revert.

cihati (changes requested):
`sdk_function` is neat, but I think making that check for each function might be a bit too much nagging.

layer/logged_data/log_data_runner.py
volkangurel: Done

cihati (approved):

layer/main/asset.py
```diff
@@ -0,0 +1,281 @@
...
+from .utils import sdk_function
+
+
+if TYPE_CHECKING:
+    import pandas
+
+logger = logging.getLogger(__name__)
+
+
+@sdk_function
```
dpp23, line 34: Would this also trigger if inside the executors? Currently users have no control over what version of the SDK is built-in the image so a warning will be confusing if they already have the latest version.

dpp23: Wouldn't it be simpler if we issue that warning only when people do `import layer`?

volkangurel: This is not changing functionality so much as moving it to a decorator. We were already checking things inside each function.

dpp23: > This is not changing functionality so much as moving it to a decorator. We were already checking things inside each function.

The bit that changed though is that before it was checking only on `layer.login` and `layer.init` which are not called inside the executors right? Now we do it for all sdk functions.

PR #316, opened 2022-08-15
Use new code execution logic for local runs

volkangurel:
This PR fixes several issues around the new code execution logic:
- Local runs
- Dataset and model status reporting
- UI tracker integration
- `layer.log` integration

Fixes LAY-3648


cihati (approved):
As I said offline: this is a very hard PR to review, due to its size & complexity. I haven't seen any issues at a glance. I approve, but I think one of @eminn / @donatasm / @maasg 's approval would be more appropriate as they know the code paths for this particular project better.

layer/clients/flow_manager.py
```diff
@@ -38,13 +38,15 @@ def start_run(
...
         user_command: str,
+        env_variables: Mapping[str, str],
     ) -> RunId:
         response = self._service.StartRunV2(
             request=StartRunV2Request(
                 project_full_name=project_full_name.path,
                 plan=execution_plan,
                 project_files_hash=Sha256(value=project_files_hash),
                 user_command=user_command,
+                env_variables=env_variables,
```
adrien-layer, line 49: this looks a bit dangerous, but i understand the need for it.

layer/contracts/definitions.py
```diff
@@ -22,6 +22,8 @@ class FunctionDefinition:
     def __init__(
         self,
         func: Callable[..., Any],
+        args: Sequence[Any],
+        kwargs: Mapping[str, Any],
```
adrien-layer, line 26: why do we need these?

layer/contracts/definitions.py
```diff
@@ -147,15 +124,58 @@ def package(self) -> Path:
...
+
+            return model_runner(self)
+        raise Exception(f"Invalid asset type {self.asset_type}")
+
+    def _clean_function_home_dir(self) -> None:
+        # Remove directory to clean leftovers from previous runs
+        function_home_dir = self.function_home_dir
+        if function_home_dir.exists():
+            shutil.rmtree(function_home_dir)
+        os.makedirs(function_home_dir)
```
adrien-layer, line 149: is this acting in a tmp location? or can it wipe the user files if they use it wrongly?

layer/decorators/layer_wrapper.py
```diff
@@ -66,12 +66,29 @@ def __init__(
...
+    def get_definition(
+        self, args: Sequence[Any], kwargs: Mapping[str, Any]
+    ) -> FunctionDefinition:
         self.layer.validate()
         current_project_full_name = get_current_project_full_name()
 
         return FunctionDefinition(
             func=self.__wrapped__,
+            args=args,
+            kwargs=kwargs,
```
adrien-layer, line 91: i find these confusing

adrien-layer (commented):
sorry i cannot approve, there is too much going on that i dont understand. i leave it to proj-simpler-execution team

layer/contracts/definitions.py
volkangurel: These are to enable https://github.com/layerai/roadmap/issues/16

layer/contracts/definitions.py
volkangurel: this is not new code, just organizing existing methods into deprecated and regular methods

layer/decorators/layer_wrapper.py
volkangurel: this is to support https://github.com/layerai/roadmap/issues/16. if args and kwargs are bound to the function via `func.bind(my_arg)`, then these are saved on the function definition and sent to the backend to be used during execution.

layer/executables/dataset/entrypoint.py
```diff
@@ -34,89 +36,69 @@
...
+    def inner() -> Any:
+        return _run(dataset_definition)
 
+    return inner
 
-def _run(user_function: Any) -> None:
-    settings: LayerSettings = user_function.layer
-    current_project_full_name_ = get_current_project_full_name()
+
+def _run(dataset_definition: FunctionDefinition) -> None:
```
adrien-layer, line 46: i think fabric should be passed down here, so it reads `_run(definition, fabric)`

i understand we get it from environment or default to `local` with
```python
def _get_display_fabric() -> str:
    return os.getenv("LAYER_FABRIC", Fabric.F_LOCAL.value)
```
but it does not mean it should be buried that deep. I prefer to inject env variables as early as possible so the rest of the code can be "normally" tested and does not care about environment.

where is the earliest we could get this and pass it down to the functions which need it (inside `run` i assume)?

layer/decorators/layer_wrapper.py
volkangurel: We have "runtimes" for this purpose defining different logic depending on where the code runs. 

The code itself is only defined once by the user, so it's captured in a single class called FunctionDefinition. The same code for function and a generic wrapper (chosen depending on `@dataset` or `@model`) is the core that doesn't change, regardless of runtime.

layer/executables/dataset/entrypoint.py
volkangurel: good point, fixing!

volkangurel: actually, this code is changed in this subsequent PR for remote runs: https://github.com/layerai/sdk/pull/322/files#diff-78e468dc28d217a4c73c94b60dee7399ec36940d901420f7d05ecf8d5147d2cf. mind if I fix it there?

layer/decorators/layer_wrapper.py
adrien-layer: ok but nonetheless here you described that args/kwargs are needed for remote execution only? or i misunderstood

hence also my comment
> i understand better now. do they have to have the same name as when we use *args: Any, **kwargs: Any in signatures or can we call them remove_execution_args, remove_execution_kwargs?

maybe we dont have a choice but to have them inside the core for binding, but then we can be explicit it is not needed in local exec?

volkangurel: sorry I didn't address that part, they are not just for remote, they are also for local execution.

volkangurel: before this PR, local execution with args worked but remote didn't, because they were using two different code paths. a big reason for this PR is to simplify this code and have them both use the same code paths for the same logic. so this args/kwargs will be used in both scenarios.

adrien-layer (approved):

PR #318, opened 2022-08-16
E2E tests for reading logged data for guests

cihati:


test/e2e/test_guest_user_reads.py
```diff
@@ -63,10 +81,23 @@ def test_guest_user_private_dataset_read(
 
         assert "not found" in str(error)
 
+        # Logged data should also not be readable.
+        dataset = client.data_catalog.get_dataset_by_name(
+            populated_project.id, "dataset1"
+        )
+
+        with pytest.raises(LayerClientException) as error:
```
adrien-layer, line 89: there is a dedicated exception for this, should be sthg like `LayerClientResourceNotFound`

test/e2e/test_guest_user_reads.py
```diff
@@ -63,10 +81,23 @@ def test_guest_user_private_dataset_read(
...
+                tag=dataset_log_tag, dataset_build_id=dataset.build.id
+            )
+        assert "not found" in str(error)
+
 
 def test_guest_user_public_dataset_read(
+    client: LayerClient,
     populated_public_project: Project,
     guest_context,
+    guest_client: LayerClient,
```
adrien-layer, line 100: why do we use both guest client and client here?

test/e2e/test_guest_user_reads.py
```diff
@@ -78,8 +109,21 @@ def test_guest_user_public_dataset_read(
         assert df.values[0][0] == "id1"
         assert df.values[0][1] == 10
 
+        dataset = client.data_catalog.get_dataset_by_name(
+            populated_public_project.id, "dataset1"
+        )
```
adrien-layer, line 114: are we using the right client here?

test/e2e/test_guest_user_reads.py
```diff
@@ -91,20 +135,49 @@ def test_guest_user_private_model_read(
         assert "not found" in str(error)
 
 
+def test_guest_user_private_model_logged_data_read(
+    client: LayerClient,
+    populated_project: Project,
+    guest_context,
+    guest_client: LayerClient,
```
adrien-layer, line 142: i understand why you have separate tests, but they do create separate runs every time which is not great for our costs and ci/test durations

can we simply have 2 tests?
1. check all private things cannot be access from guest
2. check all public things can be accessed from guest

also serves as more straight to the point documentation

test/e2e/test_guest_user_reads.py
```diff
@@ -78,8 +109,21 @@ def test_guest_user_public_dataset_read(
...
+            tag=dataset_log_tag, dataset_build_id=dataset.build.id
+        )
+
+        assert logged_data.data == str(dataset_log_value)
+        assert logged_data.logged_data_type == LoggedDataType.NUMBER
+        assert logged_data.tag == dataset_log_tag
+
 
 def test_guest_user_private_model_read(
+    client: LayerClient,
```
adrien-layer, line 126: client or guest_client?

test/e2e/test_guest_user_reads.py
adrien-layer: yes, sorry i should have spotted this earlier

adrien-layer (approved):
maybe a comment / fixture to help understand why we sometimes use non guest client

PR #326, opened 2022-08-17
Add category parameter when logging data

aleksmitov:
Extend `layer.log` with a `category` parameter, so it can be used like `layer.log({...}, category="test")`.
This will be used in the Web UI to group the logged data on it.

### DIFF
--- a/layer/clients/logged_data_service.py
+++ b/layer/clients/logged_data_service.py
@@ -63,6 +63,7 @@ class LoggedDataClient:
         tag: str,
         points: List[ModelMetricPoint],
         metric_group_id: UUID,
+        category: Optional[str] = None,
     ) -> ModelMetricId:
         metric = LoggedModelMetric(
             unique_tag=tag,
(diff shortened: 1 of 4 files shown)


cihati (approved):

fcakyon:
@aleksmitov @cihati 

In other paltforms (wandb/neptune/comet) this grouping is done via log name which makes it more structured I believe.

The charts/params are grouped in the Web UI as `layer.log({'train/accuracy: 0.5), `train/loss`, `val/accuracy` and logged items will be grouped under `train` and `val` frames.

Do you have any intention to support that instead of additional `category argument`

cihati:
@fcakyon thanks for your valuable comment!

We don't plan to add such a feature -- ultimately, we believe that having an explicit `category` argument is cleaner and more structured. One can pass a dictionary of multiple metrics and tag them as "train" with a single parameter instead of appending "train/" prefix to all of them.

PR #331, opened 2022-08-17
Fix executable remote runs

volkangurel:
Fix progress reporting and dependency installs for remote runs.

### DIFF
--- a/layer/executables/packager.py
+++ b/layer/executables/packager.py
@@ -61,16 +61,11 @@ def package_function(
             cloudpickle.register_pickle_by_value(sys.modules[function.__module__])  # type: ignore
             cloudpickle.dump(function, function_, protocol=pickle.DEFAULT_PROTOCOL)  # type: ignore
 
-        # TODO delete this once backend is updated, it's unused
-        assertions_path = source / "assertions"
-        with open(assertions_path, mode="wb") as assertions_:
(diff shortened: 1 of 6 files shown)


maasg (approved):
:heavy_check_mark:

layer/executables/packager.py
```diff
@@ -61,16 +61,11 @@ def package_function(
...
-        assertions_path = source / "assertions"
-        with open(assertions_path, mode="wb") as assertions_:
-            cloudpickle.dump([], assertions_, protocol=pickle.DEFAULT_PROTOCOL)  # type: ignore
-
         metadata_path = source / "metadata.json"
         with open(metadata_path, mode="w", encoding="utf8") as metadata_:
             json.dump(metadata or {}, metadata_, separators=(",", ":"))
 
-        target = (output_dir or Path(".")) / function.__name__
+        target = (output_dir or Path(".")) / "package.zip"
```
eminn, line 68: why hardcoded package name and I don't see any updates on how do we reference this package with the new name ?

volkangurel: This is created in a unique folder for each function anyway, so we didn't need another unique identifier that created additional complexity.

Plus, I found out when looking at the filesystem that this was practically hardcoded to "inner" for all functions because of the name of the function inside the wrapper.

PR #336, opened 2022-08-18
Add a grouped parameter to layer.log()

aleksmitov:
Add an optional boolean to `layer.log()`. If `True`, all the metrics logged with the same call to this function will be displayed on the same chart in the web UI.


volkangurel:
@aleksmitov I know we previously discussed using a boolean here, but thinking about this some more I wonder if we should add a "group_tag" string field instead. Then we can store this in the backend instead of a group UUID, and use that to refer to the graph in the UI.

volkangurel:
I'll close this PR and instead use the `group_tag` coming in a later PR.

PR #343, opened 2022-08-22
Enable executables for all non production environments

donatasm:
Enable executables for all non production environments.


maasg (approved):
:heavy_check_mark:

dpp23 (approved):

PR #347, opened 2022-08-23
Log errors but do not stop executing on logging errors (part 1)

marta-tatiana:
**Eventual Objective**: Log data in the background and do not interrupt execution if there is an error logging data. Attach error information (as a warning) to progress tracker's output in local mode.

**What this part doesn't do**: 
In this PR we're not changing any behaviour except for the added `close` call to http session when uploading files; new destination is not initialised in context yet and log function is falling back either to `ImmediateLoggedDataDestination` (which mimics current behaviour) or `ReadOnlyLoggedDataDestination` (which mimics current behaviour for the read part) based on execution context.
This PR just prepares classes so that they can be used by executables in next step, once new version of sdk is published.

**What this part does**: 
1) Introduce a concept of `LoggedDataDestination` with multiple implementations. One of them is the new implementation `QueuingLoggedDataDestination` which has new behaviour of adding logging requests to a queue, ignoring errors (if any) and just adding them to a list of error message that can be retrieved from the instance upon close.  `ImmediateLoggedDataDestination` mimics current logging behaviour with the exception that we call close on http session in `__exit__` which is not the case currently.
2) Prepare tracker to print warnings in local mode.

**Important trade-offs:**
* no support for tracker in remote mode yet, but for remote runs, the error message would be included in execution logs with the local tracker:
image

* in new queueing implementation, logging errors are not logged with python `logger` because it breaks the tracker experience in local mode. Since errors are printed below the tracker, they are be included in execution logs anyway, so I decided that additional logging would bring no benefit. 
* in new queueing implementation, when execution finishes, we wait for the reading thread for up to TIMEOUT s (currently 30s). that means that if execution is finished, and there is a single item that's still uploading for that long, we will close its http connection and give up. Other valid choice would be not to have timeout and let all uploading/retries finish no matter how long they take.

### DIFF
--- a/layer/context.py
+++ b/layer/context.py
@@ -5,6 +5,7 @@ from yarl import URL
 
 from layer.contracts.asset import AssetPath, AssetType
 from layer.contracts.datasets import DatasetBuild
+from layer.logged_data.logged_data_destination import LoggedDataDestination
 from layer.tracker.ui_progress_tracker import RunProgressTracker
 from layer.training.base_train import BaseTrain
 
@@ -45,6 +46,7 @@ class Context:
         self,
         url: URL,
         asset_path: AssetPath,
+        logged_data_destination: Optional[LoggedDataDestination] = None,
         tracker: Optional[RunProgressTracker] = None,
         train: Optional[BaseTrain] = None,
         dataset_build: Optional[DatasetBuild] = None,
@@ -63,6 +65,9 @@ class Context:
         self._asset_type = asset_path.asset_type
         self._train: Optional[BaseTrain] = train
         self._dataset_build: Optional[DatasetBuild] = dataset_build
+        self._logged_data_destination: Optional[
+            LoggedDataDestination
+        ] = logged_data_destination
         self._tracker: Optional[RunProgressTracker] = tracker
 
     def url(self) -> URL:
@@ -113,6 +118,9 @@ class Context:
 
     def asset_name(self) -> str:
         return self._asset_name
+
+    def logged_data_destination(self) -> Optional[LoggedDataDestination]:
+        return self._logged_data_destination
 
     def asset_type(self) -> AssetType:
         return self._asset_type
--- a/layer/logged_data/data_logging_request.py
+++ b/layer/logged_data/data_logging_request.py
@@ -1,0 +1,26 @@
+from typing import Any, Callable, Optional
+
+from layerapi.api.service.logged_data.logged_data_api_pb2 import LogDataResponse
+
+from layer.logged_data.file_uploader import FileUploader
+
+
+class DataLoggingRequest:
+    def __init__(
+        self,
+        files_storage: FileUploader,
+        queued_operation_func: Callable[[], Optional[LogDataResponse]],
+        data: Optional[Any] = None,
+    ) -> None:
+        self._queued_operation_func = queued_operation_func
+        self._data = data
+        self._files_storage = files_storage
+
+    def execute(self) -> None:
+        data_logging_response = self._queued_operation_func()
+        if (
+            self._data is not None
+            and data_logging_response is not None
+            and data_logging_response.s3_path is not None
+        ):
+            self._files_storage.upload(data_logging_response.s3_path, self._data)
--- a/layer/logged_data/logged_data_destination.py
+++ b/layer/logged_data/logged_data_destination.py
@@ -1,0 +1,29 @@
+from abc import ABC, abstractmethod
+from typing import Any, Callable, Optional
+from uuid import UUID
+
+from layer.clients.logged_data_service import LoggedDataClient
+from layer.contracts.logged_data import LoggedData
+
+
+class LoggedDataDestination(ABC):
+    @abstractmethod
+    def get_logged_data(
+        self,
+        tag: str,
+        train_id: Optional[UUID] = None,
+        dataset_build_id: Optional[UUID] = None,
+    ) -> LoggedData:
+        pass
+
+    @abstractmethod
+    def receive(
+        self,
+        func: Callable[[LoggedDataClient], Optional[Any]],
+        data: Optional[Any] = None,
+    ) -> None:
+        pass
+
+    @abstractmethod
+    def get_logging_errors(self) -> Optional[str]:
+        pass
--- a/layer/logged_data/queuing_logged_data_destination.py
+++ b/layer/logged_data/queuing_logged_data_destination.py
@@ -1,0 +1,111 @@
+import queue
+import threading
+from queue import Queue
+from types import TracebackType
+from typing import Any, Callable, Optional
+from uuid import UUID
+
+from layer.clients.logged_data_service import LoggedDataClient
+from layer.contracts.logged_data import LoggedData
+from layer.logged_data.data_logging_request import DataLoggingRequest
+from layer.logged_data.file_uploader import FileUploader
+from layer.logged_data.logged_data_destination import LoggedDataDestination
+
+
+WAIT_INTERVAL_SECONDS = 1
+LOGGING_TIMEOUT = 30
+
+
+class QueueingLoggedDataDestination(LoggedDataDestination):
+    def __init__(self, client: LoggedDataClient) -> None:
+        self._logged_data_client = client
+        self._files_storage = FileUploader()
+
+        self._sending_errors: str = ""
+        self._stop_reading = False
+        self._local_queue: Queue[DataLoggingRequest] = Queue()
+        self._reading_thread = threading.Thread(target=self._execute_elem_from_queue)
+
+    def __enter__(self) -> "QueueingLoggedDataDestination":
+        self._reading_thread.start()
+        return self
+
+    def __exit__(
+        self,
+        exc_type: Optional[BaseException],
+        exc_val: Optional[BaseException],
+        exc_tb: Optional[TracebackType],
+    ) -> None:
+        self._exit()
+
+    def get_logged_data(
+        self,
+        tag: str,
+        train_id: Optional[UUID] = None,
+        dataset_build_id: Optional[UUID] = None,
+    ) -> LoggedData:
+        return self._logged_data_client.get_logged_data(
+            tag=tag, train_id=train_id, dataset_build_id=dataset_build_id
+        )
+
+    def receive(
+        self,
+        func: Callable[[LoggedDataClient], Optional[Any]],
+        data: Optional[Any] = None,
+    ) -> None:
+        self._local_queue.put_nowait(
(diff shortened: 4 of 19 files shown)


layer/logged_data/data_logging_request.py
```diff
@@ -0,0 +1,20 @@
+from typing import Any, Callable, Optional
+
+from layer.logged_data.retrying_files_sender import RetryingFilesSender
+
+
+class DataLoggingRequest:
+    def __init__(
+        self,
+        files_storage: RetryingFilesSender,
+        func: Callable[[], Optional[Any]],
```
cihati, line 10: nit: let's have a better name here, smt like `queued_operation_func` or smt (same for other places where `func` is used)

layer/logged_data/queuing_network_log_destination.py
```diff
@@ -0,0 +1,78 @@
...
+        data: Optional[Any] = None,
+    ) -> None:
+        self._local_queue.put_nowait(
+            DataLoggingRequest(
+                self._files_storage, lambda: func(self._log_data_client), data
+            )
+        )
+
+    def close(self) -> None:
+        if not self._stop_reading:
```
cihati, line 38: if `self._stop_reading` is `True` and this `close` method is called, should we not do anything?

layer/logged_data/queuing_network_log_destination.py
```diff
@@ -0,0 +1,78 @@
...
+            self._stop_reading = True
+            self._reading_thread.join(timeout=3 * WAIT_INTERVAL_SECONDS)
```
cihati, line 40: Why the `3 *`?

layer/logged_data/queuing_network_log_destination.py
```diff
@@ -0,0 +1,78 @@
...
+                self._sending_errors.append(ex)
+
+    # todo: naive implementation for now, we should move this to tracker?
+    def _print_warning_message_if_errors(self) -> None:
+        if len(self._sending_errors) == 0:
+            return
+
+        message = "\nWARNING: Layer was unable to log all data because of the following errors:\n"
+        for ex in self._sending_errors:
+            message = message + "Exception: " + str(ex) + "\n"
```
cihati, line 76: nit: 
```suggestion
            message += f"Exception: {ex}\n"
```
Probably works too (haven't tried) and seems neater?

layer/logged_data/retrying_files_sender.py
```diff
@@ -0,0 +1,23 @@
+from typing import Any
+
+import requests  # type: ignore
+
+
+class RetryingFilesSender:
```
cihati, line 6: nit: I'd probably call this `FileUploader`, and rename `store` to `upload`. (I'd skip the "Retrying" part since we have one such class)

layer/logged_data/queuing_network_log_destination.py
```diff
@@ -0,0 +1,78 @@
...
+
+class QueueingNetworkLogDestination(LogDataDestination):
+    def __init__(self, client: LoggedDataClient) -> None:
+        self._log_data_client = client
+        self._files_storage = RetryingFilesSender()
+
+        self._sending_errors: List[Any] = []
+        self._stop_reading = False
+        self._local_queue: Queue[DataLoggingRequest] = Queue()
+        self._reading_thread = threading.Thread(target=self._send_elem_from_queue)
```
cihati, line 23: I don't know the answer to this, but we sometimes use `ThreadPoolExecutor` and sometimes use `Thread`. Just raising this, maybe @donatasm would chime in.

cihati (commented):
Looks good! A few comments but mostly questions. I think it's definitely worth a look from others too.

I wonder if we could utilize `contextlib.contextmanager` instead of `def close`s

marta-tatiana:
FYI reviewers - I need to adapt this PR to  @adrien-layer 's Contexts rework + we may want to print upload errors with tracker instead of standard output so there will be significant changes to everything but the contents of "LogDestination" and its implementation (which I'll rename to LoggedDataDestination ;)).

layer/context.py
```diff
@@ -81,7 +90,9 @@ def asset_type(self) -> AssetType:
             raise Exception("Unsupported asset type")
 
     def close(self) -> None:
-        pass
+        log_destination = self.log_data_destination()
```
maasg, line 93: Why do we go through the public method instead of directly accessing the instance var?

layer/logged_data/queuing_network_log_destination.py
donatasm: If single thread does the job, use `Thread`.

We use `ThreadPoolExecutor`, because it is easier to work with coroutines.

However, while `Thread` is simpler, `ThreadPoolExecutor` would have these advantages:
* increasing number of parallelism is easier
* if thread dies, executor will take care creating another
* executor is easy to change to `ProcessPoolExecutor` if we decide we need move out of proc for queue processing.

layer/logged_data/data_logging_request.py
marta-tatiana: thx, renamed

layer/logged_data/queuing_network_log_destination.py
marta-tatiana: this was significantly changed, but I kept the behavior - hopefully it now makes more sense, since error messages can be retrieved from instance multiple times

layer/logged_data/queuing_network_log_destination.py
marta-tatiana: you're right, that was a bad idea, changed to explicit timeout as described in description

layer/logged_data/queuing_network_log_destination.py
marta-tatiana: thanks, used that syntax

layer/logged_data/retrying_files_sender.py
marta-tatiana: renamed

layer/logged_data/queuing_network_log_destination.py
marta-tatiana: thank you, I sticked to `Thread`. I think it's good enough for now, and the logic is contained in one class so easy to extend/change in the future if needed.

layer/context.py
marta-tatiana: this got changed but you're right, this didn't make sense. hopefully no longer a case anywhere :)

volkangurel (approved):
Looks great overall. We'll need to make slight adjustments to the log data runner and system metrics as those are changing soon, but let's merge this in an we will iterate on top of these changes.

layer/logged_data/logged_data_destination.py
```diff
@@ -0,0 +1,29 @@
...
+
+class LoggedDataDestination(ABC):
+    @abstractmethod
+    def get_logged_data(
+        self,
+        tag: str,
+        train_id: Optional[UUID] = None,
+        dataset_build_id: Optional[UUID] = None,
+    ) -> LoggedData:
+        pass
```
cihati, line 17: This seems to have the exact same impl for all 3 subclass. Maybe extract the logic here?

layer/logged_data/queuing_logged_data_destination.py
```diff
@@ -0,0 +1,111 @@
...
+
+from layer.clients.logged_data_service import LoggedDataClient
+from layer.contracts.logged_data import LoggedData
+from layer.logged_data.data_logging_request import DataLoggingRequest
+from layer.logged_data.file_uploader import FileUploader
+from layer.logged_data.logged_data_destination import LoggedDataDestination
+
+
+WAIT_INTERVAL_SECONDS = 1
+LOGGING_TIMEOUT = 30
```
cihati, line 16: This seems too low -- we allow up to 100mb videos to be logged, for instance.

layer/logged_data/read_only_logged_data_destination.py
```diff
@@ -0,0 +1,31 @@
...
+        )
+
+    def receive(
+        self,
+        func: Callable[[LoggedDataClient], Optional[Any]],
+        data: Optional[Any] = None,
+    ) -> None:
+        raise RuntimeError("Readonly logged data destination can only be read.")
+
+    def get_logging_errors(self) -> None:
```
cihati, line 30: Might be a quirk of the typing system, but the return type is not the same as in the abstract superclass.

layer/logged_data/logged_data_destination.py
```diff
@@ -0,0 +1,29 @@
...
+
+    @abstractmethod
+    def receive(
```
cihati, line 20: this name seems quite confusing to me. Especially given the `None` return type

layer/logged_data/queuing_logged_data_destination.py
```diff
@@ -0,0 +1,111 @@
...
+        self._local_queue.put_nowait(
+            DataLoggingRequest(
+                files_storage=self._files_storage,
+                queued_operation_func=lambda: func(self._logged_data_client),
+                data=data,
+            )
+        )
+
+    def get_logging_errors(self) -> Optional[str]:
+        self._exit()
```
cihati, line 65: It feels weird to have such a side-effect for a `get` method

test/unit/logged_data/test_queuing_logged_data_destination.py
```diff
@@ -0,0 +1,52 @@
...
+        errors = destination.get_logging_errors()
+
+        # then
+        assert (
+            errors
+            == f"WARNING: Layer was unable to log requested data because of the following errors:\nException: {error_message}\nException: {other_error_message}\n"
+        )
+
+
+def test_given_logging_succesfull_then_no_errors() -> None:
```
cihati, line 43: nit: successful

test/unit/logged_data/test_queuing_logged_data_destination.py
```diff
@@ -0,0 +1,52 @@
+from typing import Any, Optional
```
cihati, line 1: It's likely hard to unit-test, but looks like there are a few untested behaviours (flushing, timeout, etc)

cihati (approved):
My comments aren't strictly blocking, feel free to merge and address in an upcoming PR.

marta-tatiana:
This was force-push due to rebase on main (there were conflicts because of https://github.com/layerai/sdk/pull/359 merged in the meantime)

marta-tatiana:
> My comments aren't strictly blocking, feel free to merge and address in an upcoming PR.

@cihati thank you, agreed on all comments, but once the checks pass I'd like very much to merge, because I'm chasing a moving target and rebasing a lot :) Proposed changes are local to logging destinations so hopefully will cause a lot less conflicts with other work.

PR #348, opened 2022-08-23
Only keep read methods in Context

adrien-layer:
We want to give users contextual methods similar to the ones available inside a decorated layer function.
The end goal is to offer:
```python

def run_experiment():
  context = layer.get_context()
  train_url = context.get_url()
  context.log({key:value})
  # or layer.log({key: value})

with layer.model("my-model"):
  run_experiment()
```
so that we are very close to Layer function executions
```python
@layer.model("my-model")
def run_experiment():
  train_url = context.get_url()
  layer.log({key:value})
```

This will help users a lot when they dont want to use decorators

## Next PRs
- [ ] do `set_active_context` and `reset_active_context` inside `__enter__` and `__exit__` instead of relying on clients to do it
- [ ] add `get_url()` to context 
-

### DIFF
--- a/layer/context.py
+++ b/layer/context.py
@@ -7,9 +7,34 @@ from layer.tracker.ui_progress_tracker import RunProgressTracker
 from layer.training.base_train import BaseTrain
 
 
+# We store the active context temporarily so it can be used within
+# either a @layer decorated function or soon via `with layer.model() as context:`
+_ACTIVE_CONTEXT: Optional["Context"] = None
+
+
+def set_active_context(context: "Context") -> None:
+    global _ACTIVE_CONTEXT
+    _ACTIVE_CONTEXT = context
+
+
+def reset_active_context() -> None:
+    global _ACTIVE_CONTEXT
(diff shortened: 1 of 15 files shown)


layer/context.py
```diff
@@ -7,20 +7,45 @@
...
+    """
+    Returns the active context object set from the active computation.
+    Used in local mode to identify which context to log resources to.
+
+    @return:  active context object
+    """
+    return _ACTIVE_CONTEXT
+
+
 class Context:
```
volkangurel, line 35: Can we add an `__enter__` and `__exit__` to this class? With those, we wouldn't need to `set_active_context` or `reset_active_context` anywhere other than this file.

volkangurel (changes requested):

layer/context.py
adrien-layer: in the description im saying i would do this in a 2nd PR :) so i limit potential conflicts in this one

for exactly this reason yes

volkangurel (approved):

PR #350, opened 2022-08-23
Fix model assertions

volkangurel:
Fix model assertions by passing in assertions for the new entrypoints vs getting it from the layer settings on the user function.


maasg:
bors try

PR #352, opened 2022-08-23
Simplify entrypoint code

volkangurel:
Simplify entrypoint code by passing in shared objects (between datasets and models) from `common.py`


layer/executables/entrypoint/common.py
```diff
@@ -16,17 +20,34 @@
...
     def runner(definition: FunctionDefinition) -> Any:
         def inner() -> Any:
             _initialize(definition)
             config: Config = asyncio_run_in_thread(ConfigManager().refresh())
-            return run_function(
-                definition, config, _get_display_fabric(), _get_run_id()
+            progress_tracker = get_progress_tracker(
+                url=config.url,
+                project_name=definition.project_name,
+                account_name=definition.account_name,
```
adrien-layer, line 38: we could start passing `ProjectFullName` in places which require both `account_name` and `project_name`

layer/executables/entrypoint/dataset.py
```diff
@@ -34,84 +32,74 @@
...
+    tracker.mark_dataset_building(dataset_definition.asset_name)
+    # TODO pass path to API instead
+    current_project_uuid = verify_project_exists_and_retrieve_project_id(
+        client, dataset_definition.project_full_name
+    )
+    dataset_build_id = client.data_catalog.initiate_build(
+        current_project_uuid,
+        dataset_definition.asset_name,
+        fabric.value,
+    )
```
adrien-layer, line 53: since `client` holds specialized clients, we _could_ move TODOs there and simplify the code here:
```python
client.initiate_dataset_build(dataset_definition.asset_path)
```
and we do this inside `client`
```python
   # TODO pass path to API instead
    current_project_uuid = verify_project_exists_and_retrieve_project_id(
        client, dataset_definition.project_full_name
    )
    dataset_build_id = client.data_catalog.initiate_build(
        current_project_uuid,
        dataset_definition.asset_name,
        fabric.value,
    )
```

adrien-layer (approved):

layer/executables/entrypoint/common.py
volkangurel: agreed, noting for a subsequent refactor PR

layer/executables/entrypoint/dataset.py
volkangurel: agreed :) noting for a subsequent refactor PR

PR #360, opened 2022-08-25
Create deletable org account in e2e test

adrien-layer:


marta-tatiana (approved):
Approved the code changes but before merging would be great to understand where did the poetry.lock change come from and if contributors are going to be overwriting this for each other or not.

PR #365, opened 2022-08-26
Remove 'EXECUTABLE_FEATURE' flag

maasg:
Removes the `LAYER_EXECUTABLES` feature flag. After this PR, all execution will go through as packaged function executed by the Executor service.


layer/config/__init__.py
```diff
@@ -14,18 +14,3 @@
...
-    if "LAYER_EXECUTABLES" in os.environ:
-        return True
-
-    try:
-        config = ConfigManager().load()
-    except:  # noqa
-        return False
-
-    # enable by default for non production environments
-    return config.url != DEFAULT_URL
```
dpp23, line 31: Can we just flip the default rather than remove completely? If a user hits issues, we can tell them to disable it as opposed to have to ship a new version in a rush.

maasg: I think that @volkangurel was eager to start deleting code.
I'd feel a bit safer flipping the flag as @dpp23 suggested. Here's that option, if we want to go for it: https://github.com/layerai/sdk/pull/367

volkangurel: I was, but I agree with @dpp23 that we can do that a bit later.

maasg:
closing in favor of https://github.com/layerai/sdk/pull/367

PR #367, opened 2022-08-26
Flip the feature flag

maasg:
Makes the function executable feature flag true by default but user override-able by explicitly setting it to `False`.

Let's chose between this option and https://github.com/layerai/sdk/pull/365 
@dpp23 @volkangurel


volkangurel (approved):

volkangurel:
Let's go with this option, so that we have an easier path to rollback for users if needed.

dpp23 (approved):

PR #370, opened 2022-08-28
Add system stats logging back to the sdk

volkangurel:
Add system stats logging back to the sdk and start tracking system metrics for all dataset and model runs.

Fixes https://github.com/layerai/roadmap/issues/12.


layer/logged_data/system_metrics.py
```diff
@@ -0,0 +1,252 @@
...
+                        f"gpu{gpu.id}_mem_used_mb": gpu.mem_used,
+                        f"gpu{gpu.id}_mem_allocated_mb": gpu.mem_total,
+                    }
+                )
+
+        return metrics
+
+
+class PsUtilMetricsCollector(MetricsCollector):
+    # cpu
```
cihati, line 64: nit: useless comment (same for `# memory` below)

cihati (commented):
So the changes under  layer/executables/entrypoint/ and layer/executables/packager.py are for fixing Marta's issues from Friday? If yes, maybe let's make it a different PR.

Also, what are the differences, if any, between the old and new system_metrics.py?

layer/logged_data/system_metrics.py
volkangurel: done!

volkangurel:
yes, I needed those changes too, so I put them in here. I can separate it out if that needs to merge in sooner.

volkangurel:
The old and new `system_metrics.py` are mostly similar. The main differences are:
- Logging data in groups so it can be visualized properly in the UI
- Logging data as time in seconds vs steps
- Small bug fixes around stepping logic
- Cleanup of logic to stop using "remote" vs "local" stats and instead use "Docker" if available otherwise fall back to PS

cihati (approved):

PR #373, opened 2022-08-29
Allow empty return for datasets

### LINKED ISSUE #244: logging experiments without a decorator and without returning model file

adrien-layer:
Fixes #244 

Already available for models, this allows logging data to Layer without sharing the underlying dataset with Layer

Screenshot 2022-08-29 at 11 33 20
Screenshot 2022-08-29 at 11 33 29
Screenshot 2022-08-29 at 11 33 35

### DIFF
--- a/layer/executables/entrypoint/dataset.py
+++ b/layer/executables/entrypoint/dataset.py
@@ -79,6 +79,17 @@ def _run(
                 result = dataset_definition.func(
                     *dataset_definition.args, **dataset_definition.kwargs
                 )
+
+            # in case client does not return the dataset
+            if result is None:
+                tracker.mark_dataset_built(dataset_definition.asset_name)
+                client.data_catalog.complete_build(
+                    dataset_build_id,
+                    dataset_definition.asset_name,
+                    dataset_definition.uri,
+                )
+                return
+
             result = check_and_convert_to_df(result)
             _run_assertions(
                 dataset_definition.asset_name,
--- a/layer/executables/entrypoint/dataset.py
+++ b/layer/executables/entrypoint/dataset.py
@@ -9,10 +9,9 @@ from yarl import URL
 from layer.clients.layer import LayerClient
 from layer.context import Context
 from layer.contracts.assertions import Assertion
-from layer.contracts.datasets import DatasetBuild, DatasetBuildStatus
+from layer.contracts.datasets import DatasetBuild
 from layer.contracts.definitions import FunctionDefinition
 from layer.contracts.fabrics import Fabric
-from layer.contracts.tracker import DatasetTransferState
 from layer.exceptions.exceptions import (
     LayerClientException,
     LayerClientServiceUnavailableException,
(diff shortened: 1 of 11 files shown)


layer/executables/entrypoint/dataset.py
```diff
@@ -74,6 +74,17 @@ def _run(
             result = dataset_definition.func(
                 *dataset_definition.args, **dataset_definition.kwargs
             )
+
+            # in case client does not return the dataset
+            if result is None:
+                tracker.mark_dataset_built(dataset_definition.asset_name)
```
volkangurel, line 80: Can we also add a `save_dataset` function similar to `save_model` to allow users save datasets on their own?

cihati (approved):

adrien-layer:
ended up refactoring a few things here to get dataset and model execution paths closer to each other.

little by little let's make it easier to understand both executions by just looking at one

PR #376, opened 2022-08-29
Lazy load ML dependencies

donatasm:
Bringing back lazy loads reverted in https://github.com/layerai/sdk/pull/356, because of pip not supporting publishing packages referencing git commit hashes.

Support for Python 3.7 was added in https://github.com/scientific-python/lazy_loader/pull/14, hence package is referenced by version now.

Not bringing back lazy loading pandas or numpy as they're always are loaded anyway, because of the [image extensions](https://github.com/layerai/sdk/blob/3a6b022982b542656bf0658a671f5b5d4af2302b/layer/__init__.py#L39).

Closes:
* https://linear.app/layer/issue/LAY-3689/lazilly-load-sdk-dependencies-that-are-not-always-needed
* https://linear.app/layer/issue/LAY-3341/import-kerascallback-and-xgboostcallback-lazily

### DIFF
--- a/test/unit/test_import.py
+++ b/test/unit/test_import.py
@@ -1,0 +1,17 @@
+from importtime_output_wrapper import (
+    get_import_time,
+    import_tree_to_waterfall,
+    parse_import_time,
+    prune_import_depth,
+)
+
+
+def test_import_layer_does_not_import_heavy_libs():
+    _, raw_output = get_import_time(import_cmd="layer", module_only=True)
+    all_imports = prune_import_depth(parse_import_time(raw_output), depth=4)
+
+    deps = import_tree_to_waterfall(all_imports, width=64)
+
+    assert "torch" not in deps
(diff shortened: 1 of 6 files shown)


test/unit/test_import.py
```diff
@@ -0,0 +1,17 @@
...
+
+def test_import_layer_does_not_import_heavy_libs():
+    _, raw_output = get_import_time(import_cmd="layer", module_only=True)
+    all_imports = prune_import_depth(parse_import_time(raw_output), depth=4)
+
+    deps = import_tree_to_waterfall(all_imports, width=64)
+
+    assert "torch" not in deps
+    assert "xgboost" not in deps
+    assert "keras" not in deps
```
cihati, line 17: nit: I'd probably assert _something expected_ to be *in* `deps` as well, to make this more sturdy.

cihati (approved):

PR #382, opened 2022-08-31
Handle both cgroups v1 and v2 for system metrics

micheldlebeau:


layer/logged_data/system_metrics.py
```diff
@@ -84,23 +103,80 @@ def get_memory_metrics(self) -> Dict[str, Dict[str, float]]:
...
+    """
+
+    def get_cpu_usage(self) -> int:
+        # Time in nanoseconds
+        return int(self._read_cgroup_metric("cpu/cpuacct.usage_user"))
+
+    def get_cpu_available(self) -> float:
+        # Times in microseconds
+        cpu_quota = int(self._read_cgroup_metric("cpu/cpu.cfs_quota_us"))
+        cpu_period = int(self._read_cgroup_metric("cpu/cpu.cfs_period_us"))
```
micheldlebeau, line 118: For all these filesystem reads, regex matches, etc. at the moment failures would crash hard.
I don't think we should crash a training simply because we can't gather metrics for some reason.
WDYT?

dpp23: I agree, but this is probably a separate ticket.

dpp23 (approved):

PR #385, opened 2022-09-01
Bundle local sdk version in executable package

dpp23:
In order to be able to be able to install the same version as the client side inside the remote executors.

In production, this means adding a file with the requirement `layer==1.2.3`, in development this means copying the local sdk code so we can install from source.

### DIFF
--- a/layer/executables/packager.py
+++ b/layer/executables/packager.py
@@ -11,7 +11,6 @@ import zipfile
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any, Callable, Dict, Generator, Optional, Sequence, Tuple
-from xmlrpc.client import Boolean
 
 import layer
 from layer.contracts.conda import CondaEnv
@@ -109,7 +108,7 @@ def get_function_package_info(package_path: Path) -> FunctionPackageInfo:
     )
 
 
-def _is_dev_version() -> Boolean:
+def _is_dev_version() -> bool:
     return layer.__version__ == "0.10.0b1"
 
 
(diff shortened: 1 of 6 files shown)


layer/executables/packager.py
```diff
@@ -11,7 +11,9 @@
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any, Callable, Dict, Generator, Optional, Sequence, Tuple
+from xmlrpc.client import Boolean
```
maasg, line 14: is this the source for the `Boolean` type that we want? xmlrpc looks odd.

layer/executables/packager.py
```diff
@@ -105,6 +109,10 @@ def get_function_package_info(package_path: Path) -> FunctionPackageInfo:
     )
 
 
+def _is_dev_version() -> Boolean:
+    return layer.__version__ == "0.10.0b1"
```
maasg, line 113: Would `__layer.__version__.endswith("0b1")` be a more stable heuristic?

maasg (approved):
:heavy_check_mark:

layer/executables/packager.py
dpp23: Silly auto import, all I needed was just `bool`. Fixed. Thank you!

layer/executables/packager.py
dpp23: I wanted this to be as strict as possible. If we ever decide to publish any beta versions, we don't want to accidentally trigger this logic.

If we change the local version, the test will fail and we can update this constant as well.

PR #386, opened 2022-09-01
Allow system metrics to work with cgroups-v2

micheldlebeau:


layer/logged_data/system_metrics.py
```diff
@@ -193,9 +253,24 @@ def __enter__(self) -> None:
...
-            self._metrics_collector = DockerMetricsCollector(self._logger)
+        path = os.path.abspath("/sys/fs/cgroup/")
+        p = [
+            p
+            for p in psutil.disk_partitions(all=True)
+            if p.mountpoint == path.__str__()
+        ]
+        cgroup_mount_type = ""
+        if len(p) == 1:
+            cgroup_mount_type = p[0].fstype
```
dpp23, line 264: Nit: Can we move that to a function with some explanation what is going on here? Then it can return an enum:

```
class CGroupsVersion(Enum):
    V1 = "v1"
    V2 = "v2"
    UNKNOWN = "unknown"
```

layer/logged_data/system_metrics.py
```diff
@@ -193,9 +253,24 @@ def __enter__(self) -> None:
             logged_data_destination=logged_data_destination,
         )
 
-        # By default, we get metrics from psutil
-        if "LAYER_FABRIC" in os.environ:
-            self._metrics_collector = DockerMetricsCollector(self._logger)
+        path = os.path.abspath("/sys/fs/cgroup/")
+        p = [
```
dpp23, line 257: Can we have a better name for this variable?

layer/logged_data/system_metrics.py
```diff
@@ -84,23 +105,86 @@ def get_memory_metrics(self) -> Dict[str, Dict[str, float]]:
...
+            return int(memory_high)
+
+
+class CGroupsVersion(Enum):
+    V1 = "v1"
+    V2 = "v2"
+    UNKNOWN = "unknown"
+
+
 class DockerMetricsCollector(MetricsCollector):
```
dpp23, line 172: Can we rename this to CGroupsMetricsCollector? I guess in the case of Ubuntu, we will be reading CGroups, but there is not necessarily docker involved there.

dpp23 (approved):

PR #387, opened 2022-09-01
Add ray as a dev dependency

almg80:
Add ray as dev dependency to allow local development to connect to a local cluster.
Type `ray start --head --port=6379` after starting a jupyter notebook and you can delegate jobs to the local cluster.

### DIFF
--- a/environment.mk
+++ b/environment.mk
@@ -72,7 +72,8 @@ else
                          $(call get_python_package_version,xgboost) \
                          $(call get_python_package_version,lightgbm) \
                          $(call get_python_package_version,h5py) \
-                         $(call get_python_package_version,pyarrow)
+                         $(call get_python_package_version,pyarrow) \
+                         $(call get_python_package_version,grpcio)
 endif
 else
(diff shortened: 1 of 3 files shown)


environment.mk
```diff
@@ -72,7 +72,8 @@ else
                          $(call get_python_package_version,xgboost) \
                          $(call get_python_package_version,lightgbm) \
                          $(call get_python_package_version,h5py) \
-                         $(call get_python_package_version,pyarrow)
+                         $(call get_python_package_version,pyarrow) \
+                         $(call get_python_package_version,grpcio)
```
dpp23, line 76: Is `gprcio==1.43` not working when installed with pip?

dpp23 (approved):

environment.mk
almg80: Not for M1, Ray errors out importing it when installed via pip on M1 linking to https://docs.ray.io/en/master/ray-overview/installation.html#m1-mac-apple-silicon-support where it asks us to conda install it.

PR #390, opened 2022-09-02
INF-1277 Test ray runtime during e2e tests

almg80:
`make ray-test` will start a local ray instance and a the new test will build a dataset and model against it.

Tests are the same as the ones that run against Layer Infra.


dpp23 (approved):
Worth checking with @eminn about the actual test code, but the setup looks good to me!

eminn (approved):

PR #393, opened 2022-09-03
Fix cgroups parsing on Ubuntu 22.04

dpp23:
On my machine, `user.slice/memory.high` seems to have an extra empty line at the end of the file so the string comparison was failing.


eminn (approved):

micheldlebeau:
>On my machine, user.slice/memory.high seems to have an extra empty line at the end of the file so the string comparison was failing.

Was it crashing the train? Or behaving differently somehow?

micheldlebeau (approved):

dpp23:
> > On my machine, user.slice/memory.high seems to have an extra empty line at the end of the file so the string comparison was failing.
> 
> Was it crashing the train? Or behaving differently somehow?

It as crashing with `I can't parse "max\n" as int`

PR #398, opened 2022-09-05
Use permanent test organization in e2e tests if it exists

adrien-layer:
We will test creation flow and access separately

Will clean up the code once all stacks can use a permanent test org account

### DIFF
--- a/test/e2e/conftest.py
+++ b/test/e2e/conftest.py
@@ -29,6 +29,9 @@ TEST_SESSION_CONFIG_TMP_PATH = DEFAULT_LAYER_PATH / "e2e_test_session.ini"
 
 TEST_ORG_ACCOUNT_NAME_PREFIX = "sdk-e2e-org-"
 
+# This organization account will be used instead of creating one if it exists
+PERMANENT_TEST_ORG_ACCOUNT_NAME = "e2e-test-org"
+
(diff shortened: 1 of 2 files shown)


test/e2e/conftest.py
```diff
@@ -173,28 +173,32 @@ async def delete_old_org_accounts() -> None:
...
-    await ConfigManager().refresh(force=True)
+    try:
+        org_account = client.account.get_account_by_name(
+            PERMANENT_TEST_ORG_ACCOUNT_NAME
+        )
+    except LayerClientResourceNotFoundException:
+        org_account_name, display_name = pseudo_random_account_name()
+        org_account = client.account.create_organization_account(
+            org_account_name, display_name, deletion_allowed=True
+        )
```
dpp23, line 188: What is the purpose of the fallback? If the org is not there, but we should probably just fail hard

adrien-layer: the purpose is to allow us to merge this before we have the test account created on all stacks

dpp23 (approved):

test/e2e/conftest.py
adrien-layer: well I should have put a better description first :) thanks!

PR #403, opened 2022-09-06
Delete e2e test projects older than one day

adrien-layer:
To account for the leaked e2e test projects

### DIFF
--- a/layer/clients/project_service.py
+++ b/layer/clients/project_service.py
@@ -106,18 +106,18 @@ class ProjectServiceClient:
             GetAccountProjectsByIdRequest(account_id=AccountId(value=str(account_id)))
         )
 
-        def should_be_deleted(p: ProjectMessage) -> bool:
+        def should_be_deleted(_project: ProjectMessage) -> bool:
             return (
-                p.created_time.ToDatetime() < query.created_before
-                and re.search(query.name_matches, p.name) is not None
+                _project.created_time.ToDatetime() < query.created_before
+                and re.search(query.name_matches, _project.name) is not None
             )
 
         deleted = 0
-        for p in list(resp.project):
-            if should_be_deleted(p):
+        for project in list(resp.project):
+            if should_be_deleted(project):
                 try:
                     if not dry_run:
-                        self.remove_project(UUID(p.id.value))
+                        self.remove_project(UUID(project.id.value))
                     deleted += 1
                 except Exception as err:
                     if not ignore_errors:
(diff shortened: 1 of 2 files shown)


layer/clients/project_service.py
```diff
@@ -81,6 +84,29 @@ def get_project_by_id(self, project_id: UUID) -> Optional[Project]:
...
+            )
+
+            def is_old_project(p: ProjectMessage) -> bool:
+                return p.created_time.ToDatetime() < date
+
+            deleted = 0
+            for p in list(resp.project):
+                if is_old_project(p):
+                    if not dry_run:
+                        self.remove_project(UUID(p.id.value))
```
dpp23, line 104: Do we want to catch the exception here, log and continue?

Otherwise there is a race between multiple runs getting the same list and then deleting orgs.

dpp23: Also can we filter test projects only?

I know some folks like @volkangurel like running tests with their own account (and I guess nothing stops an external person of doing so). We probably don't want to nuke all their projects in that case.

adrien-layer: yes good point, gonna filter on names like we do for accounts

layer/clients/project_service.py
```diff
@@ -81,6 +92,38 @@ def get_project_by_id(self, project_id: UUID) -> Optional[Project]:
...
+        Used for e2e test cleanup
+        """
+        resp: GetAccountProjectsByIdResponse = self._service.GetAccountProjectsById(
+            GetAccountProjectsByIdRequest(account_id=AccountId(value=str(account_id)))
+        )
+
+        def should_be_deleted(p: ProjectMessage) -> bool:
+            return (
+                p.created_time.ToDatetime() < query.created_before
+                and re.search(query.name_matches, p.name) is not None
```
dpp23, line 112: Nit: can we call the variable `project`? Makes it easier to read

layer/clients/project_service.py
```diff
@@ -81,6 +92,38 @@ def get_project_by_id(self, project_id: UUID) -> Optional[Project]:
...
+            )
+
+        deleted = 0
+        for p in list(resp.project):
```
dpp23, line 116:
```suggestion
        for project in list(resp.project):
```

dpp23 (approved):
A few linter nits, but otherwise looks good

layer/clients/project_service.py
adrien-layer: in such a small function there can be debate :) but since you ask, I will update it

PR #407, opened 2022-09-07
Start logging dicts as grouped logs

volkangurel:
Start logging dicts as grouped logs

### DIFF
--- a/layer/logged_data/log_data_runner.py
+++ b/layer/logged_data/log_data_runner.py
@@ -114,8 +114,11 @@ class LogDataRunner:
             elif LogDataRunner._is_tabular_dict(value):
                 if TYPE_CHECKING:
                     assert isinstance(value, dict)
-                dataframe = LogDataRunner._convert_dict_to_dataframe(value)
-                self._log_dataframe(value=dataframe, **kwargs)
+                # expand the dict into a group of logs
+                for item_key, item_value in value.items():
+                    self.log(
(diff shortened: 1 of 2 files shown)


cihati (commented):
I'll take a deeper look at this, but can you update the docs as well? This is changing existing behavior.

layer/logged_data/log_data_runner.py
```diff
@@ -114,8 +114,11 @@ def log(  # pylint: disable=too-many-statements
             elif LogDataRunner._is_tabular_dict(value):
                 if TYPE_CHECKING:
                     assert isinstance(value, dict)
-                dataframe = LogDataRunner._convert_dict_to_dataframe(value)
-                self._log_dataframe(value=dataframe, **kwargs)
+                # expand the dict into a group of logs
+                for item_key, item_value in value.items():
+                    self.log(
+                        data={item_key: item_value}, category=category, group_tag=tag
+                    )
```
cihati, line 121: @mecevit this PR reverts the behavior from https://github.com/layerai/sdk/pull/92 -- are you cool with that?

mecevit: There is no change on the UI or the SDK side, it's just backend. So I'm fine

cihati: No, I think there's a change in _both_ UI and SDK side.

PR #410, opened 2022-09-09
Refactor execution entrypoints

volkangurel:
Refactor execution entrypoints to reduce duplicate logic


layer/contracts/datasets.py
```diff
@@ -71,7 +71,6 @@ def __init__(
         self._version_id = version_id
         self.schema = schema
         self.metadata = metadata if metadata is not None else {}
-        self.build = build or DatasetBuild()
```
cihati, line 74: Why do we pass a `build` to the constructor then?

layer/clients/protomappers/datasets.py
```diff
@@ -0,0 +1,83 @@
...
+#         version=version.name,
+#         uri=build.location.uri,
+#         metadata=dict(build.location.metadata.value),
+#         build=DatasetBuild(
+#             id=build.id.value,
+#             status=DatasetBuildStatus(build.build_info.status),
+#             info=build.build_info.info,
+#             index=str(build.index),
+#         ),
+#     )
```
cihati, line 83: intentional?

test/unit/exceptions/test_exception_handler.py
```diff
@@ -1,127 +0,0 @@
...
-
-    with pytest.raises(SystemExit):
-        fut()
-
-    assert callback_instance.callback_called
-    assert callback_instance.stage == "test_2"
-    assert str(callback_instance.reason) == str(Exception("failure_2"))
-
-
-def test_exception_handler_with_system_exit() -> None:
```
cihati, line 62: Where did this test go?

test/unit/exceptions/test_exception_handler.py
```diff
@@ -1,127 +0,0 @@
...
-
-        importlib.import_module("missing_dep")
-        return
-
-    with pytest.raises(SystemExit):
-        fut()
-
-    assert callback_instance.callback_called
-    assert callback_instance.stage == "test_4"
-    assert exc_message_pattern.match(str(callback_instance.reason))
```
cihati, line 127: Why do we only test these for the model decorator now, but not for dataset?

layer/resource_manager.py
```diff
@@ -1,228 +0,0 @@
...
-            download_url, timeout=aiohttp.ClientTimeout(total=None)
-        ) as response:
-            with open(download_path, mode="wb") as f:
-                bytes_total = 0
-                async for data in response.content.iter_chunked(8192):
-                    bytes_total += f.write(data)
-        self._log.info(
-            f"Downloaded {download_url_path} to {download_path}, bytes total: {bytes_total}"
-        )
-
```
cihati, line 195: Where did this logic go?!

cihati (commented):
some comments. as I mentioned offline, it's extremely big as is. I gave some ideas to @volkangurel about how to make it smaller PRs

layer/contracts/aws.py
```diff
@@ -0,0 +1,14 @@
+from dataclasses import dataclass
+
+
+@dataclass(frozen=True)
+class AWSCredentials:
+    access_key_id: str
+    secret_access_key: str
+    session_token: str
```
adrien-layer, line 8: where do we put the types we officially support for sdk users?

SDK internals vs SDK public interfaces? are all contracts supposed to be supported? or they're simply plain python code ~ core domain inside the SDK between our APIs and User interfaces?

layer/contracts/datasets.py
```diff
@@ -25,10 +25,10 @@ class DatasetBuildStatus(enum.IntEnum):
 
 @dataclass(frozen=True)
 class DatasetBuild:
-    id: uuid.UUID = field(default_factory=uuid.uuid4)
-    status: DatasetBuildStatus = DatasetBuildStatus.INVALID
+    id: uuid.UUID
+    status: DatasetBuildStatus
+    tag: str
```
adrien-layer, line 30: the same `DatasetBuild` is also used inside `Context` but actually there status does not make sense: we are gonna need to distinguish the execution context (run/experiment) from artifacts yes (tags)

layer/contracts/models.py
```diff
@@ -18,9 +17,36 @@
...
+class ModelTrainStatus(enum.IntEnum):
+    INVALID = 0
+    PENDING = 1
+    INITIALIZING = 2
+    FETCHING_FEATURES = 3
+    IN_PROGRESS = 4
+    SUCCESSFUL = 5
+    FAILED = 6
+    CANCEL_REQUESTED = 7
+    CANCELED = 8
```
adrien-layer, line 35: why does it need to be an IntEnum and not simply an Enum?

layer/contracts/runs.py
```diff
@@ -38,6 +34,13 @@ def local_relative_paths(self) -> Iterator[str]:
                     yield os.path.relpath(dir_file_path)
 
 
+@enum.unique
+class TaskType(enum.IntEnum):
+    INVALID = 0
```
adrien-layer, line 39: why keep the invalid? let's throw when we cannot map it should be caught in E2E tests

layer/contracts/tracker.py
```diff
@@ -149,38 +149,41 @@ def __str__(self) -> str:
 
 @unique
 class AssetTrackerStatus(str, Enum):
+    """
+    PENDING -> BUILDING / TRAINING -> ASSERTING -> ASSERTED -> UPLOADING -> DONE
+    DOWNLADING -> FROM CACHE / LOADED
+    """
```
adrien-layer, line 155: how can we make sure this doesn't go obsolete? is it testable?

maasg:
Agree with @cihati 
Ideally, each goal should be in its own PR:
- Combine dataset and model logic
- Remove unused client code
- Fix bugs with progress tracking
- Start using domain objects instead of protos inside code. This will allow cleaner separation between API and SDK logic

layer/exceptions/exceptions.py
```diff
@@ -273,7 +275,7 @@ def __init__(self, message: str):
 
 
 class LayerServiceUnavailableExceptionDuringExecution(ProjectExecutionException):
-    def __init__(self, run_id: RunId, message: str):
+    def __init__(self, run_id: uuid.UUID, message: str):
```
adrien-layer, line 278: if RunId is an important concept, we could introduce one in `contracts` then if using `uuid.UUID` instead reduces clarity and understandability of the code?

layer/logged_data/system_metrics.py
```diff
@@ -298,12 +298,12 @@ def __enter__(self) -> None:
...
-        train = active_context.train()
-        train_id = train.get_id() if train is not None else None
+        model_train = active_context.model_train()
+        model_train_id = model_train.id if model_train is not None else None
         dataset_build = active_context.dataset_build()
         dataset_build_id = dataset_build.id if dataset_build is not None else None
         self._log_data_runner = LogDataRunner(
-            train_id=train_id,
+            train_id=model_train_id,
             dataset_build_id=dataset_build_id,
```
adrien-layer, line 307: this part we can generalize further in the future i feel

layer/contracts/datasets.py
volkangurel: I believe this is a remnant from an old piece of code that is no longer used

layer/clients/protomappers/datasets.py
volkangurel: nope, good catch, removed!

test/unit/exceptions/test_exception_handler.py
volkangurel: We're going to merge more and more of models and datasets together (in a subsequent PR). For now, testing these for models should be sufficient as datasets go through almost the exact same codepath.

layer/resource_manager.py
volkangurel: It is no longer used since resources are embedded into the executable zip now

layer/contracts/aws.py
volkangurel: haven't thought through that part yet :) but I do like the idea of having separate classes for publicly supported objects vs internal domain objects. maybe a task for a future PR.

layer/contracts/datasets.py
volkangurel: yes, absolutely. that is hopefully coming soon

layer/contracts/models.py
volkangurel: this is based off of the DatasetBuildStatus for consistency.

layer/contracts/runs.py
volkangurel: agreed, fixed (in https://github.com/layerai/sdk/pull/414)

layer/contracts/tracker.py
volkangurel: it's more for internal documentation than anything else. i found that this logic existed in a different part of the codebase so I wanted to surface it here. I can't think of an easy way of testing this.

layer/exceptions/exceptions.py
volkangurel: yes, we can do that in the next iteration of adding domain objects

layer/logged_data/system_metrics.py
volkangurel: absolutely. we may also just switch over to "artifact_id" and stop using dataset build and model train

layer/executables/entrypoint/dataset.py
```diff
@@ -1,172 +1,67 @@
...
-                result,
-                dataset_definition.assertions,
-                tracker,
+        if self.run_id:
+            self.client.flow_manager.update_run_metadata(
+                run_id=self.run_id,
+                task_id=self.definition.asset_path.path(),
+                task_type=TaskType.DATASET_BUILD,
+                key="build-id",
+                value=str(self.build.id),
```
adrien-layer, line 42: feels like this could also be in common but because all parameters differ almost i understand why it remains here.

this part though is common and important for remote execution, while being not explicit enough at a higher level that we're in a Layer cloud execution environment
```python
if self.run_id:
  self.client.flow_manager.update_run_metadata(
```

layer/executables/packager.py
```diff
@@ -32,8 +32,8 @@ def package_function(
 ) -> Path:
     """Packages layer function as a Python executable."""
 
-    if not inspect.isfunction(function) or function.__name__ == "<lambda>":
-        raise ValueError("function must be a function")
+    if not callable(function):
```
adrien-layer, line 35: nice one, didnt know it

layer/projects/progress_tracker_updater.py
```diff
@@ -116,14 +117,23 @@ def _handle_task_succeeded(self, task: PBTask) -> None:
         task_name = self._get_name_from_task_id(task_id)
         task_type = task.type
         if task_type == PBTask.TYPE_DATASET_BUILD:
+            dataset_build_id = uuid.UUID(
+                self.run_metadata[(task_type, task_id, "build-id")]
+            )
```
adrien-layer, line 122: can we have a `run_metadata.get_dataset_build_id()` instead?

or are we too close to the API here? from the name `progress_tracker_updater` it seems we're dealing with protos way to much in this class

layer/projects/progress_tracker_updater.py
```diff
@@ -116,14 +117,23 @@ def _handle_task_succeeded(self, task: PBTask) -> None:
...
+
+            build = self.client.data_catalog.get_build_by_id(dataset_build_id)
             self.tracker.mark_done(
                 asset_type=AssetType.DATASET,
                 name=task_name,
+                tag=build.tag,
             )
         elif task_type == PBTask.TYPE_MODEL_TRAIN:
+            train_id = uuid.UUID(self.run_metadata[(task_type, task_id, "train-id")])
```
adrien-layer, line 131: same can we have a `run_metadata.get_model_train_id()` instead?

adrien-layer (approved):
non-blocking comments, let's merge this so we reduce possibility of conflicts there after

PR #413, opened 2022-09-09
[LAY-3768] Rewire error reporting for Dataset build

maasg:
Application-level errors resulting from the Dataset Build process are reported to the metadata store.
We want to display those specific errors to the user in a concise way. 
This PR "rewires" the Dataset Build result into the error reporting, falling back onto the job log in case we don't have a status.
(which could be the case for OOM errors, for example)

Model train will be done separately because we need an [extra endpoint in the backend](https://linear.app/layer/issue/LAY-3770/model-catalog-should-offer-a-getmodeltrainstatus-endpoint) to use the same method exploited here.

### DIFF
--- a/layer/clients/data_catalog.py
+++ b/layer/clients/data_catalog.py
@@ -324,6 +324,12 @@ class DataCatalogClient:
         )
         return dataset_proto_mapper.from_dataset_build(build, version)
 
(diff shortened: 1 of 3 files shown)


aleksmitov (approved):
🚀

PR #416, opened 2022-09-11
Fix logging file uploads for queued logs

volkangurel:
Fix logging file uploads for queued logs. This fixes an issue that caused videos to not render in the web UI

Fixes LAY-3766

### DIFF
--- a/layer/logged_data/queuing_logged_data_destination.py
+++ b/layer/logged_data/queuing_logged_data_destination.py
@@ -1,3 +1,4 @@
+import pathlib
 import queue
 import threading
 from queue import Queue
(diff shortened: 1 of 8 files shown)


layer/logged_data/queuing_logged_data_destination.py
```diff
@@ -40,12 +41,14 @@ def receive(
         self,
         func: Callable[[LoggedDataClient], Optional[Any]],
         data: Optional[Any] = None,
+        data_path: Optional[pathlib.Path] = None,
     ) -> None:
```
cihati, line 45: Maybe throw if both `data` and `data_path` are given?

cihati (approved):

layer/logged_data/queuing_logged_data_destination.py
volkangurel: Done!

PR #422, opened 2022-09-15
Submit workflow to the created cluster

eminn:


layer/clients/executor_service.py
```diff
@@ -42,3 +48,21 @@ def get_function_download_path(
...
+        idle_scale_down_duration = Duration()
+        idle_scale_down_duration.FromTimedelta(td=idle_scale_down_duration_td)
+        idle_shut_down_duration = Duration()
+        idle_shut_down_duration.FromTimedelta(td=idle_shut_down_duration_td)
+        # TODO(emin): add account_id and language_version as parameter to below
+        request = GetOrCreateClusterRequest(
+            idle_scale_down_duration=idle_scale_down_duration,
+            idle_shut_down_duration=idle_shut_down_duration,
+            max_number_of_workers=10,  # TODO(emin): remove this parameter
+            worker_fabric="",  # TODO(emin): remove this parameter as it should be defined by the function executable for each
```
eminn, line 66: Until we figure out how do we make them configurable and expose it to the user we can keep them as defaults.

layer/projects/ray_workflow_project_runner.py
```diff
@@ -63,6 +74,13 @@ def run(self) -> Run:
         ray.shutdown()
         return run
 
+    def _get_ray_proxy_url(self) -> str:
```
yuranos, line 77: Proxy URL is what I return from `getOrCreateCluster`. You don't need to add prefix or anything else to it I think.

yuranos (approved):

layer/projects/ray_workflow_project_runner.py
eminn: That bit would be changed to return the ray head node service name so that would be added to the header for proper routing. In this method we are building the URL for the ray proxy from the connected stack URL.

layer/projects/ray_workflow_project_runner.py
```diff
@@ -63,6 +74,13 @@ def run(self) -> Run:
...
+        from urllib.parse import urlsplit, urlunsplit
+
+        url = list(urlsplit(str(self._config.url)))
+        url[1] = f"ray.{url[1]}"
+        return urlunsplit(url)
```
dpp23, line 82: I don't this is what we need there, but I'm adding a `rayGatewayDomain` to `/__details.json` which we can just fetch the same way we fetch the auth0 URL.

PR #425, opened 2022-09-16
Start a run for each execution

volkangurel:
Start a run for each execution and add the run ids to dataset builds and model trains. The next step after this is to migrate existing entities to runs and start logging data with run ids.


Makefile
```diff
@@ -22,7 +22,7 @@ $(INSTALL_STAMP): pyproject.toml poetry.lock $(PREREQ_STAMP)
 ifdef IN_VENV
 	$(POETRY) install
 else
-	$(POETRY) install --remove-untracked
+	$(POETRY) install --sync
```
adrien-layer, line 25: what does this change? -> is it worth a comment?

layer/clients/run_service.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def get_run_by_id(self, run_id: UUID) -> Run:
+        request = GetRunByRunIdRequest(
+            run_id=run_proto_mapper.to_run_id(run_id),
+        )
+        run_pb = self._service.GetRunByRunId(request=request).run
+        return run_proto_mapper.from_run(run_pb)
+
+    def get_run_by_index(self, project_id: UUID, run_index: int) -> Run:
+        raise NotImplementedError()
```
adrien-layer, line 37: 🎏  in case you forgot this one

layer/clients/run_service.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def create_run(self, project_id: UUID, run_name: Optional[str] = None) -> Run:
+        request = CreateRunRequest(
+            project_id=project_proto_mapper.to_project_id(project_id),
+        )
+        if run_name:
+            request.name = run_name
```
adrien-layer, line 44: should it be optional? the UI might be harder to deal migrate if name is optional. Also, current users are used to `layer.model("my-model")` for running experiments.

Do runs have indexes within a project? Or for a given project and run_name?

Trying to determine is optional run name is a necessary complexity

layer/contracts/definitions.py
```diff
@@ -80,32 +68,20 @@ def _get_source(self) -> str:
...
-    @property
-    def project_full_name(self) -> ProjectFullName:
-        return ProjectFullName(
-            account_name=self.account_name,
-            project_name=self.project_name,
-        )
-
     @property
     def asset_path(self) -> AssetPath:
+        project_full_name = context.get_project_full_name()
```
adrien-layer, line 73: why do they need to be taken out of constructor / properties?

invoking `context` from deep within like will create headaches down the road imo
-> said another way: is there really a situation when the `FunctionDefinition` is already done (instantiated), and we change the account/project before invoking it?

Why dont we use `context.get_project_full_name()` at creation time?

layer/contracts/definitions.py
```diff
@@ -128,11 +104,11 @@ def runner_function(self) -> Any:
         if self.asset_type == AssetType.DATASET:
             from layer.executables.entrypoint.dataset import DatasetRunner
 
-            return DatasetRunner(self)
+            return DatasetRunner(self, context.get_run_context())
```
adrien-layer, line 107: same question here, why not ask for context in 
```python
def runner_function(self, context: Context) -> Any:
```
?

it will become hard to unit test because we're hiding dependency injection when we're invoking context outside of constructors

layer/runs/initializer.py
```diff
@@ -0,0 +1,121 @@
...
+    """
+
+    def __init__(self, layer_client: LayerClient):
+        self._layer_client = layer_client
+
+    def setup_project(
+        self,
+        project_full_name: ProjectFullName,
+        run_name: Optional[str] = None,
+        run_index: Optional[int] = None,
```
adrien-layer, line 34: i thought `index` was auto-incremented by the BE? why is it an argument?

layer/tracker/ui_progress_tracker.py
```diff
@@ -62,11 +59,12 @@ def track(self) -> Iterator["RunProgressTracker"]:
             self._progress.refresh()
 
     def _get_url(self, asset_type: AssetType, name: str) -> URL:
+        project_full_name = context.get_project_full_name()
```
adrien-layer, line 62: same comment here: is there no way to pass one thing explicitly? if passing `project_full_name` does not make sense anymore, maybe context does?

it could be that it is the best way, Im just very cautious about `context` because it hides dependencies and side effects a lot. 

At least let's use it implicitly only for reading like here, not for mutating it

adrien-layer (commented):
Mainly questioning the multiple needs to invoke context without passing it as a dependency, due to maintainability concerns

Makefile
volkangurel: this is to remove a deprecation warning printed by `poetry`, since `remove-untracked` was renamed in v 1.2.0. The functionality has not changed.

cc @dpp23 for visibility

layer/clients/run_service.py
volkangurel: it should be optional, @C-McIntyre has designs to fall back to the creation time if the name is not given. we're trying to make it even more simple to run an experiment, so the fewer things to require the better IMHO.

runs have indexes within a project, names are just strings to make it easier to view things in the UI, but are not otherwise required for anything.

layer/contracts/definitions.py
volkangurel: good point, i think there is no use case where context changes once `FunctionDefinition` is initialized, so we can simplify things that way.

there is some other refactoring that I'd like to do for `FunctionDefinition` as well, once we switch over to using functions for remote execution only (as opposed to using it for experiment tracking and artifact storage along with remote execution). mind if we leave those changes till then?

layer/contracts/definitions.py
volkangurel: agreed, that can be passed in. will update!

layer/runs/initializer.py
volkangurel: it's an argument in case a user wants to resume an existing run. this was indeed your suggestion the other day 😄

layer/tracker/ui_progress_tracker.py
volkangurel: agreed, I think we should pass in the RunContext here.

PR #426, opened 2022-09-21
Label runs

adrien-layer:
Needs https://github.com/layerai/monorepo/pull/6795 first for e2e-tests to pass


layer/clients/label.py
```diff
@@ -45,6 +49,18 @@ def add_labels_to(
         )
         self._service.AddLabelsToEntity(request=request)
 
+    def add_labels_to_run(
+        self,
+        run_id: UUID,
+        label_names: Set[str],
+    ) -> None:
+        request = AddLabelsToRunRequest(
+            run_id=RunId(value=str(run_id)),
```
volkangurel, line 58: You can use the proto mapper here: https://github.com/layerai/sdk/blob/main/layer/clients/protomappers/runs.py#L11

layer/clients/label.py
```diff
@@ -65,3 +81,13 @@ def get_labels_attached_to(
             self._service.GetLabelsAttachedToEntity(request=request)
         )
         return set(map(lambda l: l.name, resp.label))
+
+    def get_labels_attached_to_run(
+        self,
+        run_id: UUID,
+    ) -> Set[str]:
+        request = GetLabelsAttachedToRunRequest(run_id=RunId(value=str(run_id)))
```
volkangurel, line 89: Same as above you can use the proto mapper: https://github.com/layerai/sdk/blob/main/layer/clients/protomappers/runs.py#L11

volkangurel (approved):
Approving with some non-blocking comments

PR #428, opened 2022-09-21
Adapt executor api changes 

eminn:
- Refresh configuration as well in case of refresh
- Add grpc channel credentials

### DIFF
--- a/layer/clients/executor_service.py
+++ b/layer/clients/executor_service.py
@@ -3,12 +3,14 @@ from datetime import timedelta
 from typing import Tuple
 
 from google.protobuf.duration_pb2 import Duration
+from layerapi.api.ids_pb2 import AccountId
 from layerapi.api.service.executor.executor_api_pb2 import (
     GetFunctionDownloadPathRequest,
     GetFunctionUploadPathRequest,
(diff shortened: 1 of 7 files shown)


layer/clients/executor_service.py
```diff
@@ -63,6 +65,12 @@ def get_or_create_cluster(
             idle_scale_down_duration=idle_scale_down_duration,
             idle_shut_down_duration=idle_shut_down_duration,
             max_number_of_workers=10,  # TODO(emin): remove this parameter
-            worker_fabric="",  # TODO(emin): remove this parameter as it should be defined by the function executable for each
+            worker_fabric="f-small",  # TODO(emin): remove this parameter as it should be defined by the function executable for each
+            account_id=AccountId(value=str(account_id)),
```
volkangurel, line 69: nit: we can use an "account_proto_mapper" here instead of creating the AccountId directly

volkangurel (approved):
Small non-blocking comment, please feel free to merge in!

layer/clients/executor_service.py
eminn: Will address on the coming ones as there are TODOs as well

